PR #1, opened 2016-01-11
Added licensing

swertz:
- Added license & headers for MoMEMta (GPLv3)
- Correct copying information for external libraries (boost, lua, spdlog, cuba)


blinkseb:
Nice :+1:

PR #3, opened 2016-01-14
Add support for travis

blinkseb:
Build time is about 5 minutes since we need to build also lhapdf on the worker node.


swertz:
Great, thanks!

PR #6, opened 2016-02-01
Create an install target for cmake

blinkseb:
The public headers files are now in 'include/momemta', while the private
ones are in 'src/'. Users will only be able to include public files into
their code, since only these headers will be installed.

This is needed to allow MoMEMta to be used in other softwares, and also to pave the way to modular inclusion of ME into the code.


swertz:
Thanks! Wouldn't it be better to have the private headers and the sources in separate directories?

blinkseb:
Yes why not. I actually have found both method recommended so it's up to you :)

swertz:
With the present number of files in src/ it's fine but if it increases too much in the future it might become messy, no?

blinkseb:
I agree. What about:

```
include/momemta
momemta/src
momemta/include
```
?

Or we can even move private headers back into `include` while letting public ones into `include/momemta`?

swertz:
So, following the discussion online, the conclusion was to have
```
include/momemta
core/src
core/include
```
Right?

blinkseb:
Yes indeed! I'll update the PR asap

blinkseb:
It's done! I've moved everything in `core/`. Much cleaner now thanks!

swertz:
Thanks!

PR #8, opened 2016-04-06
Load matrix element dynamically

blinkseb:
I'm pretty happy with the current implementation, so I open this PR for comments and review.

This changeset allow MoMEMta to load matrix element from dynamic library. Each matrix element must implement the [`MatrixElement` interface](https://github.com/blinkseb/MoMEMta/blob/external_me/include/momemta/MatrixElement.h), and register itself in our Matrix Element factory, using the `REGISTER_MATRIX_ELEMENT` macro. A minimal working matrix element can be found [here](https://github.com/MoMEMta/MoMEMta/blob/963859f6b572ba45af41e5f41d64eedf4ff754f1/MatrixElements/dummy/dummy_me.cc).

A name is assigned to each matrix element, chosen by the user when calling the `REGISTER_MATRIX_ELEMENT`. This name identifies uniquely the matrix element. The `MatrixElement` module now expect the user to set the matrix element name to use inside the configuration file ([see here](https://github.com/MoMEMta/MoMEMta/commit/0efea037e87da5900b052e90351dbb70d117fa8a#diff-3b550af74eb5b0e7b57aabff3b05ac50R119) and [here](https://github.com/MoMEMta/MoMEMta/commit/e9c2783d105b769eb279df47fa4ccf027c5aeeb4#diff-52673c82e3767514b9f9ccfff30219a5R585) for the definition of the matrix element named `pp_tt`).

Matrix elements stored into dynamic library can be loaded from the configuration file using the `load_modules` lua method. Once executed, all the matrix elements in the library are available to use from the `MatrixElement` block.

There's also some improvements of the code not directly related to the matrix elements.

---

TODO before merging:

 - [x] Decide if we use a namespace for the public interface or not. Currently, all the classes are inside a `momemta` namespace to avoid any clash, but maybe it's not needed.
 - [x] Rename the matrix element from `pp_tt` to `pp_ttx_fully_leptonic`
 - [ ] ~~Rename the `load_modules` lua method to something more generic (`load_plugins`?)~~ (postponed)


blinkseb:
On my side, it's ready to be merged once the tests are successful :)

PR #9, opened 2016-04-06
Configure CUBA from configuration file

blinkseb:
This PR allows most of the CUBA configuration to be changed from the configuration file. A default set of option is provided, and can be override from the configuration file.

Note: PR based on #8, only the last 3 commits are new


include/momemta/Utils.h
```diff
@@ -206,4 +206,29 @@ void apply_permutations(std::vector<T>& vec, std::vector<std::size_t> const& p)
 
 std::string demangle(const char* name);
 
+template <typename T> void UNUSED(T &&) { }
```
swertz: What is this for?

swertz:
Very nice, thanks :+1:!

We can take it as it is now; at some point we'll have to think about how to handle the gridfile, see the options at https://github.com/MoMEMta/MoMEMta/pull/9/commits/24b32c491d5c23cbb383ce9f331271cd6003abe2#diff-8503a635c4b36c07fdea9f4442ca9bbaL121 which are fixed for now...

Also, it would be good if these options could be modified using functions in the user's code, not only from the config file. But that's not urgent!

include/momemta/Utils.h
blinkseb: It's just to trick the compiler to think arguments or variable are used, even if it's not the case. Since the function does nothing, it'll be optimized away when turning on optimization so it has no impact on runtime, but remove a lot of warnings :)

blinkseb:
Yep indeed. I chose to let only _vital_ options to be configurable for the moment, but it should be rather easy to support the grid options.

Concerning the ability to change the parameters from the code, it is indeed a must-have feature. But unfortunately, it's not so easy to do, especially with the current way of parsing the configuration...

include/momemta/Utils.h
swertz: I see :p Of course (some of) these variables which are unused now should also be passed to the integrand, as one can imagine use cases where modules would need to acess them, but it's not important for the moment...

vidalm:
Great man !!

PR #11, opened 2016-04-07
Add doxygen documentation support

blinkseb:
Preliminary support for doxygen. When running `cmake`, a new target `docs` is added if doxygen is detected. To create the documentation, simply run

```
make docs
```

and magic should occurs. The documentation will be located in `documentation/html`. I also added a small doxygen theme to make the HTML output prettier :blush:


vidalm:
So, we should drop .md files in docs and that's it !!
Thanks

PR #12, opened 2016-04-08
Added permutator module

swertz:
This PR introduces the new module "Permutator" to handle Monte-Carlo integration of the permutations of identical particles. 

It adds a dimension of integration, and using this extra phase-space point, decides which permutation to use to permutate the entries in the "input" table to produce the "output" table.

In the example config, add an option "USE_PERM" to decide if the module is used or not. If it is, it takes the two input b-jets (which either come out of the TFs or come directly from the user if no TF is used) and handles the permutation. Its output is then passed with the rest (i.e., the leptons) to the BlockD module.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,7 @@ set(MOMEMTA_SOURCES
     "modules/Flatter.cc"
     "modules/GaussianTransferFunction.cc"
     "modules/MatrixElement.cc"
+    "modules/Permutator.cc"
     "core/src/ConfigurationReader.cc"
     "core/src/ConfigurationSet.cc"
     "core/src/InputTag.cc"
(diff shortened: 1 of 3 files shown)


blinkseb:
Very nice thanks!

Can you check the indentation please? It should be 4 spaces ;)

Also, we currently do *not* use the boost library, isn't the included header available in C++11?

swertz:
I fixed the indentation. Regarding boost, this function is not available in STL, and the boost header is not standalone: http://www.boost.org/doc/libs/1_59_0/boost/math/special_functions/factorials.hpp.

swertz:
Changed to TMath::Factorial, and reverted changes in the example main.cc.

vidalm:
Beloved Root

swertz:
Well, actually we could do without the factorial altogether (thanks @blinkseb ) ;) ...

blinkseb:
:+1: for me! Thanks :)

PR #13, opened 2016-04-08
Block B

vidalm:
Block B only, it needs to be tested, even just checking the phase space volume as suggested by @swertz.

### DIFF
--- a/modules/BlockB.cc
+++ b/modules/BlockB.cc
@@ -58,7 +58,7 @@ class BlockB: public Module {
             // From eq.(1) p1z = -B*E1 + A
             // From eq.(4) + eq.(1) (1-B^2)* E1^2 + 2AB* E1 - C = 0    
  
-            const double A = (s12-p22+2*(pT.Px()*p2.Px()+pT.Py()*p2.Py()))/(2*p2.Pz());
+            const double A = (s12 - p22 + 2*(pT.Px()*p2.Px() + pT.Py()*p2.Py()))/(2*p2.Pz());
             const double B = p2.E()/p2.Pz();
             const double C = SQ(pT.Px()) + SQ(pT.Py());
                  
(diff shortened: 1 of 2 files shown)


modules/BlockB.cc
```diff
@@ -0,0 +1,118 @@
...
+        double computeJacobian(const LorentzVector& p1, const LorentzVector& p2) {
+          
+            const double E1  = p1.E();
+            const double p1z = p1.Pz();
+
+            const double E2  = p2.E();
+            const double p2z = p2.Pz();
+            // Some extra info in MadWeight Source/MadWeight/blocks/class_b.f Good luck!!
+            
+            double inv_jac = 4.*SQ(M_PI*sqrt_s)*( p2z*E1 - E2*p1z);
```
swertz: Here it would be good to check that inv_jac is not too close to zero, as is done in blockD (taken from madweight)... But it's kind of ad-hoc, not great...

swertz:
Thanks! Looks good... You just forgot to add the source file in the CMakeLists ;)

blinkseb:
:+1: for me, thanks :) Indeed you also need to add the file in CMakeLists.txt in order to build it :)

vidalm:
Ok, added to CMakeLists.txt

blinkseb:
Not building. See error log: https://travis-ci.org/MoMEMta/MoMEMta/builds/121774792

modules/BlockB.cc
```diff
@@ -0,0 +1,118 @@
...
+            // FIXME
+            const LorentzVector ISR; // = (*particles)[0];
+            auto pT = p2 + ISR;
+            
+            const double p22 = p2.M2();
+            
+            // From eq.(1) p1z = -B*E1 + A
+            // From eq.(4) + eq.(1) (1-B^2)* E1^2 + 2AB* E1 - C = 0    
+ 
+            const double A = (s12 - p22 + 2*(pT.Px()*p2.Px() + pT.Py()*p2.Py()))/(2*p2.Pz());
```
blinkseb: `s12` here is not a `double`, but a `shared_ptr<double>`, so it won't work. Simply replace `s12` by `*s12` to make it compiles :)

swertz:
FYI, if you're just fixing small things in the last commit, you can do "git commit -a --amend", save the file without editing, and then "git push origin (your_branch)" to overwrite the last commit ;)

vidalm:
Thanks!! I am learning a lot today.

PR #14, opened 2016-04-11
Created 'examples' dir and moved stuff there

swertz:
As mentioned on slack, this is just to move stuff to "examples" directory, to separate the core of the code (library) from the small executable actually calling the functions and loading the config file.

Again: it's a suggestion, if you have strong feelings about this don't hesitate!


blinkseb:
You also need to edit the path to the configuration file which is hardcoded in `main.cc`

swertz:
Oops, thanks, done!

blinkseb:
It seems it's the correct next move. We should have one **full** configuration file for each Matrix Element we ship, even with some P4 hardcoded, to check that everything is still running and produce the expected result.

PR #16, opened 2016-04-12
Binned transfer functions

swertz:
This adds the module "BinnedTransferFunctionOnEnergy".

It's a first step, with no attempt to check the type of the particle (ie, should be the same for every event). Also, one could imagine having only one module for all the transfer functions on different variables (angles, ...) that can be described by a 2D histogram, where the user would configure in the lua file which transformation the module should apply to the LorentzVector, to avoid code bloat of duplicating some of the code to many TF modules (to be discussed).

I changed the output "jacobian" in the gaussian TF module to "TF_times_jacobian", to be more explicit.

I added an example for the module in the tt_fullyleptonic lua file, but commented it since it needs to access a ROOT file stored on ingrid.

There is some rudimentary documentation in the new file, but we'll surely have to settle for a common structure for all the modules and update this one.

This PR includes #14.


modules/BinnedTransferFunctionOnEnergy.cc
```diff
@@ -0,0 +1,170 @@
...
+
+        virtual size_t dimensions() const override {
+            return 1;
+        }
+
+        // We assume TF=0 for Egen < lower limit of the TH2, and
+        //           TF=constant for Egen > upper limit of the TH2
+        // In the former case, the integrated range is adapted (shortened) to the range where TF!=0
+        // TODO: create picture for documenting the TFs
+        inline double GetDeltaRange(const double &Erec) const {
```
blinkseb: Should be private I guess

modules/BinnedTransferFunctionOnEnergy.cc
```diff
@@ -0,0 +1,170 @@
...
+            return GetDeltaMax(Erec) - m_deltaMin;
+        }
+
+        inline double GetDeltaMax(const double &Erec) const {
```
blinkseb: Same here

modules/BinnedTransferFunctionOnEnergy.cc
swertz: I don't know if it should, but it can ;)

swertz:
It can be merged but please don't do it yet, we're waiting for the documentation scripts to be ready to check with this PR that they work as intended...

PR #17, opened 2016-04-12
Automatically deploy documentation on new commit

blinkseb:
This PR enable auto-deployment of the documentation on each new commit done on the `prototype` branch by travis.

If the conditions are met (a commit, not a PR), the following happens:
 - First, the SSH key is decrypted (`travis/decrypt_key.sh`). It's safe since only travis own the private key for decrypting the SSH key.
 - The documentation is built if the build succeed (and it should, since we only do PR with validation).
 - The documentation is pushed automatically to the `gh-pages` of the `MoMEMta` repository.

Documentation is then accessible using this url: http://momemta.github.io/MoMEMta/ (empty for the moment)

It's not easy to test if it's working correctly because you need a commit on the branch to trigger the build, and it work only on this repository (decryption of the key is forbidden by travis if the PR is done by a fork for security reasons). So, in a perfect world, once this PR is merged, it should trigger a documentation build! Let's see!


swertz:
Great, let's see what it gives :)

PR #18, opened 2016-04-13
Travis and documentation updates

blinkseb:
Various fixed and updates for travis and the documentation:

 - Use precompiled binaries for lhapdf instead of building our own version each time (~3 minutes gained on the total build time)
 - Fix and update to the documentation rendering


swertz:
Test time down from 6 to 3 minutes :+1:

Let's see if the doc renders as expected :)

PR #19, opened 2016-04-13
Use a graph to represent modules dependencies

blinkseb:
The graph also us to quickly:
 - detect non-existing inputs
 - remove modules with non-used outputs
 - detect cyclic dependencies (currently not done, but can be easily added)
 - order module execution based on which depend on what

### DIFF
--- a/core/src/Graph.cc
+++ b/core/src/Graph.cc
@@ -1,0 +1,157 @@
+#include <Graph.h>
+#include <logging.h>
+
+#include <momemta/Module.h>
+
+#include <boost/graph/graphviz.hpp>
+#include <boost/graph/topological_sort.hpp>
+
+namespace graph {
+
+class unresolved_input: public std::runtime_error {
+    using std::runtime_error::runtime_error;
+};
+
+Graph build(const Pool::DescriptionMap& description, std::vector<ModulePtr>& modules, std::function<void(const std::string&)> on_module_removed) {
+
+    Graph g;
+
+    uint32_t id = 0;
+    std::unordered_map<std::string, vertex_t> vertices;
+
+    // Create vertices. Each vertex is a module
+    for (const auto& d: description) {
+        vertex_t v = boost::add_vertex(g);
+        vertices.emplace(d.first, v);
+
+        g[v].name = d.first;
+        g[v].id = id;
+        auto module = std::find_if(modules.begin(), modules.end(), [&d](const ModulePtr& m) { return m->name() == d.first; });
+        if (module != modules.end())
+            g[v].module = *module;
+        id++;
+    }
+
+    modules.clear();
+
+    // Create edges. One edge connect one module output to one module input
+    for (const auto& vertex: vertices) {
+        const auto& outputs = description.at(vertex.first).outputs;
+
+        // Find all modules having for input this output
+        for (const auto& output: outputs) {
+
+            for (const auto& module: description) {
+                // Skip ourself
+                if (module.first == vertex.first)
+                    continue;
+
+                for (const auto& input: module.second.inputs) {
+                    if ((input.module == vertex.first) && (input.parameter == output)) {
+                        edge_t e;
+                        bool inserted;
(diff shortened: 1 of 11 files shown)


include/momemta/impl/InputTag_fwd.h
```diff
@@ -0,0 +1,95 @@
...
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
swertz: Why did you have to split this header in two and move it to a subfolder?

include/momemta/Pool.h
```diff
@@ -24,41 +24,66 @@
...
         template<typename T> std::shared_ptr<const T> get(const InputTag& tag) {
             if (tag.isIndexed()) {
                 throw std::invalid_argument("Indexed input tag cannot be passed as argument of the pool. Use the `get` function of the input tag to retrieve its content.");
             }
 
-            auto it = m_pool.find(tag);
-            if (it == m_pool.end())
-                throw tag_not_found_error("No such tag in pool: " + tag.toString());
+            auto it = m_storage.find(tag);
+            if (it == m_storage.end())
```
swertz: Shouldn't you check here that the pool is not frozen?

include/momemta/impl/InputTag_fwd.h
blinkseb: It's because there's a circular dependency between `Pool` and `InputTag`. The `Pool` now only includes the class definition of `InputTag` (`InputTag_fwd.h`), while all others classes includes directly `InputTag.h`, which is just a shortcut for `impl/InputTag_fwd.h` and `impl/InputTag.h`. This works because the `Pool` only needs the declaration, and does not actually call any templated function of the `InputTag`.

include/momemta/Pool.h
blinkseb: Indeed, but it's mentioned in the documentation that the freeze of the `Pool` is not enforced for the moment. I want to see how things evolve (we'll probably need in the future to access and/or put things in the `Pool` after the creation step.).

I can however easily add the check if it's wanted.

core/src/Graph.cc
```diff
@@ -0,0 +1,157 @@
...
+};
+
+void graphviz_export(const Graph& g, const std::string& filename) {
+
+    std::ofstream f(filename.c_str());
+
+    auto vertices_name = boost::get(&Vertex::name, g);
+    auto edges_name = boost::get(&Edge::name, g);
+
+    boost::write_graphviz(f, g, make_label_writer(vertices_name), make_label_writer(edges_name), boost::default_writer(), boost::get(&Vertex::id, g));
```
swertz: Where is `make_label_writer` defined?

core/src/Graph.cc
```diff
@@ -0,0 +1,157 @@
...
+
+    // Re-fill modules vector with new content (sorted & cleaned)
+    for (const auto& vertex: sorted_vertices) {
+        modules.push_back(g[vertex].module);
+    }
+
+    return g;
+}
+
+class edge_writer {
```
swertz: Where is this used?

core/src/Graph.cc
blinkseb: It's a global boost function

core/src/Graph.cc
blinkseb: Actually nowhere... But it'll be needed if we decide to colorize the output of the graphviz graph. I started to do it, but forgot about it...

core/src/Graph.cc
swertz: Ok, strange it's not in the `boost` namespace...!

PR #22, opened 2016-04-14
Markdown file describing Block B

vidalm:
New .md file added to /docs describing Block B. Let's see how this works.


swertz:
So to summarize our offline discussion: you're adding this to check how Markdown files get included in the docs, and the plan for the blocks would be to have short description in the module's .cc file (with the inputs, outputs, parameters), but a more complete description with the all math in Markdown files...

Any comments on this plan?

blinkseb:
A global comment: closing tag for Doxygen math is `\f$` and not `$\f`. The later won't work I'm afraid.

vidalm:
Ah but this is much better, we can replace everywhere $$ in Latex by \f$. Fixed

docs/BlockB.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Invisible particle, with four momenta \f$p_1\f$ (up to two solutions possible)
+- Jacobian, one per solution.    
+
+### System of equations to compute \f$p_1\f$
+
+The integrator throws random points in the mass (\f$s_{12}\f$) while \f$p_2\f$ is a known quantity. The equations to 
+compute \f$p_1\f$ are:
+
+- (1) \f$(p_1 + p_2)^2 = s_{12} = M_{1}^{2} + M_{2}^2 + 2 E_1 E_2 + 2 p_{1x}p_{2x} + 2p_{1y}p_{2y} + p_{1z}p_{2z}\f$
```
swertz: Here you could try using a numbered list? See https://en.wikipedia.org/wiki/Markdown#Example

docs/BlockB.md
```diff
@@ -0,0 +1,64 @@
+# Block B description
+Note: Block B, also referred as final block (FB), and class in the MadWeight world.
```
swertz: Irony? ;)

docs/BlockB.md
vidalm: Done.

docs/BlockB.md
vidalm: Real life !!! :-)

PR #25, opened 2016-04-19
NWA module

swertz:
Add a module to handle the narrow width approximation for a propagator one does not want to integrate over.

In practice the blocks are used the same way, but one replaces a `Flatter` module by a `NarrowWidthApproximation` module as input for a block's `s` invariant.

Added a minimal working example to demonstrate the use of this module without cloggering the other example.

Note: works, but not validated physics-wise (as all the rest).


blinkseb:
Wha, awesome!! Thanks a lot :+1:

PR #26, opened 2016-04-19
Fixed typos in docs

swertz:
Fixed a couple of issues with Latex, and experiment the non-inline styles for equations: \f[ ... \f] and \f{eqnarray}{ ... }.


blinkseb:
Let's see! Thanks :+1:

PR #27, opened 2016-04-20
PSvol test for blockD [not correct!]

swertz:
Compute the phase-space volume for blockD (6 final massless particles) assuming sqrt(s) = 1 TeV. According to MW, it should be "694 GeV^6". I don't get that number for the moment, but using this PR's content we have something to play with and debug (could also be a simple unit conversion problem?)...

In more details:
- Modify the "dummy" ME to always return 1
- Modify the MatrixElement module to be able to choose not to use the PDFs (using a non-mandatory boolean parameter)
- Add a "tests" subdirectory where we will put all our unit tests
- In this subdirectory, add a .cc and a config file for the phase space volume computation for blockD. Note: the "top" and "W" masses and widths are rubbish, but it's just so that the integration has an easier job of integrating the whole phase-space (if the widths are too small, the convergence is just too small).
- Add modules: "flat" transfer functions on |P|, Theta and Phi, allowing to integrate the whole phase space. Then, the 4-vectors passed as argument to computeWeights() should have no impact on the final answer (note: their mass **does** have an impact - in the example .cc, they are massless).
- Add a module which generates a phase-space point uniformly in a user-set range. I though I would need that but in the end didn't... Since it might still be useful sometimes, I leave it there :p
- Also, fix (again) a typo in the LaTeX of the doc


modules/FlatTransferFunctionOnP.cc
```diff
@@ -0,0 +1,100 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <momemta/ConfigurationSet.h>
+#include <momemta/Module.h>
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+
+#include <logging.h>
```
blinkseb: Header not used

modules/FlatTransferFunctionOnPhi.cc
```diff
@@ -0,0 +1,90 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <momemta/ConfigurationSet.h>
+#include <momemta/Module.h>
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+
+#include <logging.h>
```
blinkseb: Same here

modules/FlatTransferFunctionOnTheta.cc
```diff
@@ -0,0 +1,91 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <momemta/ConfigurationSet.h>
+#include <momemta/Module.h>
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+
+#include <logging.h>
```
blinkseb: And here :p

swertz:
Indeed :p

CMakeLists.txt
```diff
@@ -91,8 +91,12 @@ set(MOMEMTA_SOURCES
     "modules/BlockD.cc"
     "modules/Boost.cc"
     "modules/Flatter.cc"
+    "modules/UniformGenerator.cc"
     "modules/GaussianTransferFunction.cc"
     "modules/BinnedTransferFunctionOnEnergy.cc"
+    "modules/FlatTransferFunctionOnP.cc"
+    "modules/FlatTransferFunctionOnPhi.cc"
+    "modules/FlatTransferFunctionOnTheta.cc"
     "modules/NarrowWidthApproximation.cc"
```
blinkseb: Can you please order the files alphabetically? :blush:

CMakeLists.txt
```diff
@@ -159,6 +164,9 @@ target_link_libraries(example_tt_fullyleptonic_NWA momemta)
 set_target_properties(example_tt_fullyleptonic_NWA PROPERTIES OUTPUT_NAME
   "tt_fullyleptonic_NWA.exe")
 
+# Test executables
```
blinkseb: It would be great to have a flag to disable the compilation of the tests in case it's not needed

swertz:
Done! Call `cmake .. -DTESTS=ON` to compile the tests (default off).

PR #28, opened 2016-04-20
 Add support for delayed lua evaluation and configuration edition

blinkseb:
A lazy value is an expression whose evaluation is delayed. In our context,
it essentially means a value which can be changed after the configuration
file is parsed.

An new lua function is added, `parameter` which accepts a string argument.
Instead of referencing directly the variable in the lua configuration, the
user must use this new function, and pass as argument the name of the
parameter, as specified in the `configuration` table (see updated lua
configuration for an example)

It's also now possible to edit parameters directly from the code instead of the configuration file. The new philosophy is now:

 1. You open a lua configuration file using `ConfigurationReader`
 2. You (may) edit the parameters you want
 3. You *freeze* the reader and receive a cloned version of the configuration that you can no longer edit.
 4. You use the frozen version of the configuration to create an instance of MoMEMta.
 5. Go back to step 2 if you want

--

TODO before merging:

 - [x] Ensure that a given parameter actually exists in the `configuration` table when evaluating the Lazy function (currently give a cryptic message)
 - [x] Support editing of a given parameter after the evaluation of the configuration file

### DIFF
--- a/core/include/lua/utils.h
+++ b/core/include/lua/utils.h
@@ -25,6 +25,10 @@ namespace lua {
         using std::runtime_error::runtime_error;
     };
 
+    class unsupported_type_error: public std::runtime_error {
+        using std::runtime_error::runtime_error;
+    };
+
     /**
      * \brief List of all supported lua types
      */
@@ -39,14 +43,29 @@ namespace lua {
     };
 
     /**
-     * \brief Lazy value in lua
+     * \brief Lazy value in lua (delayed evaluation)
      *
-     * It's just a small wrapper around a lua anonymous function.
-     *
-     * Use Lazy::operator()() to evaluate the function
+     * Use Lazy::operator()() to evaluate the value
      */
     struct Lazy {
         lua_State* L; ///< The global lua state. This state must be valid for as long as this instance.
+
+        Lazy(lua_State* L);
+
+        /**
+         * \brief Evaluate the lazy value
+         *
+         * \return The value after evaluation
+         */
+        virtual boost::any operator() () const = 0;
+    };
+
+    /**
+     * \brief Lazy function in lua (delayed function evaluation)
+     *
+     * It's just a small wrapper around a lua anonymous function.
+     */
+    struct LazyFunction: public Lazy {
         int ref_index; ///< The reference index where the anonymous function is stored.
 
         /**
@@ -54,7 +73,7 @@ namespace lua {
          *
          * \return The return value of the anonymous function
          */
-        boost::any operator() () const;
+        virtual boost::any operator() () const override;
 
         /**
          * \brief Bind a anonymous lua function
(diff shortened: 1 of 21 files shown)


blinkseb:
I've updated the code a bit. It's now needed to `freeze` the configuration before passing it to MoMEMta. Next step is an interface to actually edit a parameter before freezing the configuration.

blinkseb:
Pull-requested updated. On my side it seems to be complete. Editing of parameters work from the C++ code, but it's only limited to scalar type of the moment. It should be enough for the moment, and I hope it'd be rather easy to add the remaining type later.

core/include/lua/utils.h
```diff
@@ -3,54 +3,263 @@
...
+     * to the top of the stack.
+     *
+     * \param L the current lua state
+     * \param value The value to convert
+     *
+     * \warning Vectors are currently not supported
+     */
+    void push_any(lua_State* L, const boost::any& value);
 
     template<typename T> T special_any_cast(const boost::any& value) {
```
swertz: Why is this needed?

blinkseb: It's used when parsing vectors, to provide implicit conversion from integer to double (see specialization of the template in the .cc file)

core/src/ConfigurationSet.cc
```diff
@@ -48,7 +58,83 @@ void ConfigurationSet::parse(lua_State* L, int index) {
...
+
+    return lua::to_any(L, index);
+}
+
 bool ConfigurationSet::exists(const std::string& name) const {
     auto value = m_set.find(name);
     return (value != m_set.end());
 }
+
+void ConfigurationSet::setInternal(const std::string& name, Element& element, const boost::any& value) {
```
swertz: Why is the name needed here? It's also not used in LazyConfigurationSet...

core/src/ConfigurationSet.cc
```diff
@@ -48,7 +58,83 @@ void ConfigurationSet::parse(lua_State* L, int index) {
...
+
+    element.value = value;
+    element.lazy = false;
+}
+
+void ConfigurationSet::freeze() {
+    if (frozen)
+        return;
+
+    frozen = true;
```
swertz: Maybe move this after the for loop, since it's not true if it gets interrupted (and someone catches the exceptions, who know...)?

core/src/ConfigurationSet.cc
blinkseb: I thought it may be useful in the future in we add new subclasses

core/src/ConfigurationSet.cc
blinkseb: If an exception is thrown during freezing, we would be half-frozen: some ConfigurationSet would be frozen, while some won't be. I honestly don't know how we can recover from such a state, so I don't think moving the frozen state after the for loop will change anything, but I can do it if you want :)

swertz: No it's fine ;)

swertz:
Seems very good to me! A very, very big thanks for that!

Just a question before merging: why do you say it's limited to the scalar type? https://github.com/MoMEMta/MoMEMta/pull/28/files#diff-233043c4c174e0e51fc0ae56a94c4ceaR75 would also work with a vector right?

vidalm:
I am lost, but I believe you !!!

blinkseb:
Yes indeed, but the only thing the user can change are instances of `LazyConfigurationSet`, so this line: https://github.com/MoMEMta/MoMEMta/pull/28/files#diff-233043c4c174e0e51fc0ae56a94c4ceaR139

which call this method: https://github.com/MoMEMta/MoMEMta/pull/28/files#diff-d3b5c19707aaf6852add40ca17648b40R285

and there's no support for vector. It's possible to add, but it's a bit of work because you need to create the corresponding lua table from the vector.

swertz:
Thanks for clearing that up!

PR #29, opened 2016-04-21
added BlockF

alesaggio:
Config file removed


alesaggio:
It should be ok now!

blinkseb:
Thanks a lot!

PR #30, opened 2016-04-21
add blockF

alesaggio:
documentation adjusted


swertz:
Thanks! Unfortunately there are conflicts... Have you rebased against the prototype branch?

To check that the doc is correct this time, you can try building it on ingrid using `~sbrochet/.local/bin/doxygen Doxyfile` and then visualising it using `firefox documentation/html/index.html`. Should work but never tried :D

PR #31, opened 2016-04-26
Unit tests for lua parser

blinkseb:
Nothing really interesting here, but at least it's a start for unit testing. I added the catch framework for unit-test (https://github.com/philsquared/Catch) and started to test the lua parser. It's super easy to add new tests, so if you want to contribute, please go ahead: in a perfect world, each function should be tested :)

These tests are ran by travis automatically in order to catch any regressions as soon as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,6 @@ before_script:
   - cd build
   - cmake .. -DTESTS=ON
 script:
-  - make -j2
+  - make -j2 && ./tests/unit_tests/unit_tests.exe
 after_success:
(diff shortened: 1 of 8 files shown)


swertz:
Thanks a lot!

So, we agree that we separate the unit tests from other tests in the https://github.com/MoMEMta/MoMEMta/tree/prototype/tests directory in that unit tests are ran at each push, but other tests serve as one-time validations? Since computing the PS volumes or cross sections is really slow, it's not something we can do on a regular basis, but should still check from time to time that it's not broken...

blinkseb:
That's what I had in mind, especially if the phase-space integration takes one week :laughing:

blinkseb:
PR rebased

PR #32, opened 2016-04-27
Check solutions in blockD, B and F; add test for cross-section

swertz:
- Do not consider the PS points in the blocks if they are not physical (e.g. s12 > sqrt_s*sqrt_s)
- Check that solutions are physical in the blocks
- Add a test computing the cross section of dileptonic ttbar (blockD), to compare with MadGraph.  Result of the test: 6.69 +- 0.03 pb with MoMEMta, and 6.73 +- 0.02 pb MadGraph.


blinkseb:
Thanks a lot!

PR #33, opened 2016-04-27
Build cuba using 64-bits integers

blinkseb:
Title says everything. This allows to use more than 2 147 483 647 iterations, which seems to be necessary to compute the phase-space volume.

Credits go to @swertz!


swertz:
Thanks!

PR #34, opened 2016-04-27
Define CubaIndex() function to generate cuba PS point input tag

swertz:
Instead of using `cuba::ps_point/i` as input tag in the config file, where the user has to be sure `i` is always correctly incremented and unique, introduce the function `CubaIndex()`, which always returns the correct input tag.

`examples/tt_fullyleptonic.lua` was modified to illustrate this behaviour.

### DIFF
--- a/core/src/lua/utils.cc
+++ b/core/src/lua/utils.cc
@@ -452,6 +452,12 @@ namespace lua {
         lua_pushlightuserdata(L, ptr);
         lua_pushcclosure(L, parameter, 1);
         lua_setglobal(L, "parameter");
+
+        // Define the `CubaIndex()` function in Lua and make it available in the global namespace.
+        // See generate_cuba_index for more information.
+        lua_pushnumber(L, 0);
+        lua_pushcclosure(L, generate_cuba_index, 1);
+        lua_setglobal(L, "CubaIndex");
     }
 
     std::shared_ptr<lua_State> init_runtime(IOnModuleDeclared* callback) {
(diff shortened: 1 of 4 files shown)


core/src/ConfigurationReader.cc
```diff
@@ -32,6 +32,11 @@ ConfigurationReader::ConfigurationReader(const std::string& file) {
     LOG(debug) << "Parsing LUA configuration from " << file;
     lua_state = lua::init_runtime(this);
 
+    // Define the `CubaIndex()` function in Lua and make it available in the global namespace.
+    // See generate_cuba_index for more information.
+    lua_pushcfunction(lua_state.get(), lua::generate_cuba_index);
+    lua_setglobal(lua_state.get(), "CubaIndex");
+
```
blinkseb: Please move this to the `init_runtime` function

core/src/lua/utils.cc
```diff
@@ -468,4 +468,22 @@ namespace lua {
         return L;
 
     }
+
+    int generate_cuba_index(lua_State* L) {
+        int n = lua_gettop(L);
+        if (n != 0) {
+            luaL_error(L, "invalid number of arguments: 1 expected, got %d", n);
```
blinkseb: I guess it should be `0` instead of `1`?

blinkseb:
Looks nice thanks! Would it be possible to add an unit test for this function in `tests/unit_tests/lua.cc`?

I'm also not sure if the static variable is the right thing to do. If you try to parse two configuration files, the second won't have the right indices... The counter should be reset to 0 at some point at the end of the parsing.

swertz:
Updated!

swertz:
I also added a small unit test for this new function.

blinkseb:
Thanks a lot for the test!

PR #35, opened 2016-04-28
Renaming

blinkseb:
Perform some renaming before first release:

 - `vegas` -> `cuba`
 - `ConfigurationSet` -> `ParameterSet` (the biggest one)
 - MatrixElement module output is now `integrands` and no longer `weights`
 - `CubaIndex` is renamed to `getpspoint` (all lowercase to respect Lua convention)

--

 - [x] Flatter
 - [x] Rename phase-space module's inputs
 - [x] Rename CubaIndex


swertz:
Thank you! I was thinking if there was anything else...?
There is some confusion between Cuba and Vegas, the MatrixElement module is actually more than just the ME, the flatter could be renamed "BreitWignerGenerator", the modules' phase-space points don't always expect the same input tag (sometimes "ps_point", sometimes "input" -> rename everything to "ps_point"?), the function "CubaIndex" I added yesterday actually doesn't return an index (-> "GetPSPoint"?)...

blinkseb:
Sure, let's go for the full renaming :) I'll update the PR asap

vidalm:
Yes the ME module has a confusing name. Yesterday we said Integrant, but not sure.

swertz:
Thanks! There is also some mixing between snake_case and camelCase... Should we try to choose one and stick to it?

blinkseb:
Good on my side

swertz:
Great! We'll make another PR to fix the indexing in lua then?

PR #36, opened 2016-04-28
Fix possible memory leak

blinkseb:
Leak found while running clang static analyze (awesome tool by the way :+1:)

--
~~Based on #35, to be rebased before merging~~


blinkseb:
PR rebased

PR #37, opened 2016-04-28
Start modules documentation

blinkseb:
A beginning at documenting modules and organizing a bit the documentation. All the modules are now listed in a `Modules` group so you can see at once what's available.

I did not document every modules, only the `Module` class, so any help is welcome in the future :)

--

~~PR based on #36, to be rebased~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,6 @@
 # MoMEMta
 Modular Matrix Element Method implementation
+
+## Documentation
+
+- @subpage modules "List of all available modules"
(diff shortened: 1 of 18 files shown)


vidalm:
I think we have to assign Module->person to document. We have to share the load other wise it won't work. Anybody doing the list of homework?

blinkseb:
PR rebased

swertz:
See #38

PR #40, opened 2016-05-02
Documenting  GaussianTransferFunction.cc

vidalm:
First try. As I said to Seb. W. I am not convince of the first paragraph. It is a copy paste from the Seb's binned TF. I do not have any other good idea to rephrased it (vacation mode) so...


blinkseb:
More than enough IMO. We thanksfully don't write a novel :stuck_out_tongue_closed_eyes:

modules/GaussianTransferFunction.cc
```diff
@@ -26,7 +26,34 @@
...
+ * and evaluates the transfer function on the "reconstructed" and "generated" energies.
+ *
+ * The transfer function (TF) is a Gaussian distribution that describes the difference between 
+ * the reconstructed and the generated energy (\f$E_{rec}-E_{gen}\f$). The width of the distribution depends 
+ * on the "reconstructed" energy and it is set as parameter. 
+ *
+ *  - Integration dimension: 1
+ *
+ *  - Parameters:
+ *    - `sigma` (double): Fraction of the "reconstructed" energy corresponding to the sigma of the Gaussian distribution.
```
swertz: I'm sorry but I don't find this definition very clear :p

modules/GaussianTransferFunction.cc
```diff
@@ -26,7 +26,34 @@
...
-/**
+/** \brief Transfer function on energy described by a Gaussian distribution
+ *
+ * This module takes as inputs a LorentzVector and a phase-space point, generates
+ * a new LorentzVector with a different energy (keeping direction and invariant mass),
+ * and evaluates the transfer function on the "reconstructed" and "generated" energies.
+ *
+ * The transfer function (TF) is a Gaussian distribution that describes the difference between 
+ * the reconstructed and the generated energy (\f$E_{rec}-E_{gen}\f$). The width of the distribution depends 
+ * on the "reconstructed" energy and it is set as parameter.
```
swertz: I was just thinking, in principle the width should depend on the generated energy right? The resolution is usually expressed as a function of the true energy?

modules/GaussianTransferFunction.cc
blinkseb: isn't `width of the gaussian` enough?

swertz: Well, the width is really `sigma*E`, right?

modules/GaussianTransferFunction.cc
vidalm: In principle yes, but it is hard to implement in this way. Actually this Gaussian TF is just to play. The right way would be a fitted TF where you know this info.

swertz: True... There might be a way, but let's keep that for a more realistic TF module.

PR #41, opened 2016-05-02
Add preliminary support for DMEM

swertz:
Make some changes to make things such as DMEM possible:
- Add virtual method `leafModule` to the `Module` class, to flag modules whose products are not used as input to any other module: when the graph is built, do not remove modules which return `true`.
- The module producing the integrand doesn't have to be the last module in the config file anymore: the output `integrands` is now reserved for this purpose (as a consequence, if no module, or more than one module produces it, it is an error).
- Add the "virtual" input tag `cuba::ps_weight` to have access the the phase space point's weight.
- `Pool::put` is now templated to allow calling any constructor for objects stored in the pool.
- Add the method `MoMEMta::getPool` to allow the user to retrieve objects in the pool after the integration is over.
- Add the virtual methods `atBeginIntegration` and `atEndIntegration` to the `Module` class, to be called right before or after the integration for each event.
- Add the DMEM module, which does... DMEM, for total invariant mass only... I'm not sure if we can avoid defining a module for every variable we want to plot...

I added a little bit to the example for tt to show how to use the new DMEM module. 

I still want to add some more documentation, check that the output distributions are correct, and maybe rename things to more explicit. Until then, comments are welcome!

### DIFF
--- a/core/src/MoMEMta.cc
+++ b/core/src/MoMEMta.cc
@@ -38,6 +38,7 @@ MoMEMta::MoMEMta(const Configuration& configuration) {
     // Create phase-space points vector, input for many modules
     m_pool->current_module("cuba");
     m_ps_points = m_pool->put<std::vector<double>>({"cuba", "ps_points"});
+    m_ps_weight = m_pool->put<double>({"cuba", "ps_weight"});
 
     // Create vector for input particles
     m_pool->current_module("input");
@@ -61,9 +62,23 @@ MoMEMta::MoMEMta(const Configuration& configuration) {
     // Resize pool ps-points vector
     m_ps_points->resize(m_n_dimensions);
 
-    // The last module of the chain *must* output a vector of integrands, used for the integration
+    // Find which modules produces the 'integrands' output, which defines the integrand.
+    // **There can be only one!**
     m_pool->current_module("momemta");
-    m_integrands = m_pool->get<std::vector<double>>({m_modules.back()->name(), "integrands"});
+    bool foundIntegrands = false;
+    for (const auto& module: m_modules) {
+        if(m_pool->exists({module->name(), "integrands"})) {
+            if(!foundIntegrands){
+                m_integrands = m_pool->get<std::vector<double>>({module->name(), "integrands"});
+                foundIntegrands = true;
+                LOG(debug) << "Module " << module->name() << " produces the integrand.";
+            } else {
+                throw integrands_output_error("Only one module can produce the `integrands` output.");
+            }
+        }
+    }
+    if(!foundIntegrands)
+        throw integrands_output_error("No module found which produces the mandatory `integrands` output.");
 
     m_cuba_configuration = configuration.getCubaConfiguration();
 
@@ -89,7 +104,15 @@ MoMEMta::~MoMEMta() {
     }
 }
 
+const Pool& MoMEMta::getPool() const {
+    return *m_pool;
+}
+
 std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector<LorentzVector>& particules) {
+    
+    for (const auto& module: m_modules) {
+        module->beginIntegration();
+    }
 
     *m_particles = particules;
 
@@ -141,16 +164,21 @@ std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector
          &error,                 // (double*) integration error ([ncomp])
          &prob                   // (double*) Chi-square p-value that error is not reliable (ie should be <0.95) ([ncomp])
     );
+    
+    for (const auto& module: m_modules) {
+        module->endIntegration();
+    }
 
     return std::vector<std::pair<double, double>>({{mcResult, error}});
 }
 
 double MoMEMta::integrand(const double* psPoints, const double* weights) {
 
-    UNUSED(weights);
-
     // Store phase-space points into the pool
     std::memcpy(m_ps_points->data(), psPoints, sizeof(double) * m_n_dimensions);
+
+    // Store phase-space weight into the pool
+    *m_ps_weight = *weights;
 
     for (auto& module: m_modules) {
         module->work();
--- a/include/momemta/Module.h
+++ b/include/momemta/Module.h
@@ -53,6 +53,12 @@ class Module {
          * \brief Called once at the beginning of the job
          */
         virtual void configure() { };
+        
+        /**
+         * \brief Called once at the beginning of the integration
+         */
+        virtual void beginIntegration() {};
+        
         /**
          * \brief Main function
          *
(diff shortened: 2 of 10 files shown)


core/src/MoMEMta.cc
```diff
@@ -61,9 +62,24 @@ MoMEMta::MoMEMta(const Configuration& configuration) {
     // Resize pool ps-points vector
     m_ps_points->resize(m_n_dimensions);
 
-    // The last module of the chain *must* output a vector of integrands, used for the integration
+    // Find which modules produces the 'integrands' output, which defines the integrand.
+    // **There can be only one!**
     m_pool->current_module("momemta");
-    m_integrands = m_pool->get<std::vector<double>>({m_modules.back()->name(), "integrands"});
+    bool foundIntegrands(false);
```
blinkseb: `bool foundIntegrands = false;` please

core/src/MoMEMta.cc
```diff
@@ -89,7 +105,15 @@ MoMEMta::~MoMEMta() {
     }
 }
 
+Pool& MoMEMta::getPool() {
```
blinkseb: We should return a const here, we don't want the user to change the Pool. Something like `Pool const& MoMEMta::getPool() const;`

core/src/MoMEMta.cc
```diff
@@ -141,17 +165,22 @@ std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector
...
 double MoMEMta::integrand(const double* psPoints, const double* weights) {
 
-    UNUSED(weights);
-
     // Store phase-space points into the pool
     std::memcpy(m_ps_points->data(), psPoints, sizeof(double) * m_n_dimensions);
 
+    // Store phase-space weight into the pool
+    *m_ps_weight = *weights;
+
```
blinkseb: Is `weights` a vector?

core/src/MoMEMta.cc
swertz: I though about it, but the problem is that https://github.com/MoMEMta/MoMEMta/blob/prototype/include/momemta/Pool.h#L49 is not marked const...

include/momemta/Pool.h
```diff
@@ -70,6 +70,15 @@ class Pool {
...
         void alias(const InputTag& from, const InputTag& to);
 
+        /** \brief Check if input tag exists in the pool.
+         *
+         * \warning No check is performed to ensure that the string is an input tag. Use InputTag::isInputTag() first.
+         *
+         * \param tag Input tag that will be searched in the pool.
+         * \return True if the input tag exists in the pool, False otherwise.
+         */
+        bool exists(const InputTag& tag);
```
blinkseb: This method should be `const`

modules/DMEM.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <momemta/Module.h>
+
+class DMEM: public Module {
+    public:
+
+        DMEM(PoolPtr pool, const ParameterSet& parameters): Module(pool, parameters.getModuleName())
+        {
+            xStart = parameters.get<double>("xStart");
+            xEnd = parameters.get<double>("xEnd");
+            nBins = parameters.get<int64_t>("nBins");
```
blinkseb: For the 3 parameters, please use `xstart`, `xend` and `nbins` to match lua convention.

core/src/MoMEMta.cc
swertz: In our case it's not supposed to be...? It's a vector of size https://github.com/MoMEMta/MoMEMta/pull/41/files#diff-8503a635c4b36c07fdea9f4442ca9bbaR149.

core/src/MoMEMta.cc
blinkseb: Indeed. Can you try to make it `const`? You should only have to mark the `m_description` variable as `mutable`, otherwise it should work.

core/src/MoMEMta.cc
blinkseb: Ok great. Something to keep in mind in the future is we decide to throw a bunch of points at the same time.

swertz: Yup. We're not there yet, but would you prefer replacing this "1" with a variable, and using a `memcpy` as well for the weights?

blinkseb: Oh no it's fine like this, I was just wondering is the number of weights was the same as the number of phase-space points. Since it's not, it's fine like it is :+1:

core/src/MoMEMta.cc
swertz: I'll try (as soon as ingrid is back up, I'd take me some time to setup everything on my pc), but would that be desirable? Before freezing `get` would not be const at all, isn't that misleading?

Of course in practice the user has only access to the pool through the MoMEMta instance, meaning that the pool is already frozen and the `const` is valid...

blinkseb: `get` is always `const`: you get a pointer to a `const` object. The rest is implementation details to deal with the fact that a module can be declared before another.

I agree in this end it won't change much, but I prefer to avoid crazy user who put things to the Pool between event, which we allow if we don't return a const.

If it's too much work, we can do it later on another PR, but IMO this has to be done.

include/momemta/Module.h
```diff
@@ -53,6 +53,12 @@ class Module {
          * \brief Called once at the beginning of the job
          */
         virtual void configure() { };
+        
+        /**
+         * \brief Called once at the beginning of the integration
+         */
+        virtual void atBeginIntegration() {};
```
blinkseb: Can it be renamed `beginIntegration()`? Same for `endIntegration()`.

blinkseb:
Overall it looks very good, thanks a lot. I just have one concern about the variadic template in `put`: if a module first use `get` before `put`, there's no way to reserve correct class instance on the Pool, and will see some really weird stuff. I'm wondering if there's an elegant solution to this problem, expect throwing if a user call `put` with some argument while a `get` already happened?

blinkseb:
One can use `sizeof...(T)` to know the size of the variadic template, so it's at least possible to implement the check mentioned above.

core/src/MoMEMta.cc
swertz: To make it work I also had to mark `m_storage` as `mutable`, and `create()` as `const` (kind of unnatural?).
Actually, since `put()` is private, the user cannot add anything to the pool through `MoMEMta::getPool` anyway, right? On the other hand, it was already possible before, just by putting a `produce()` in `Module::work()`...

swertz:
PR updated, but see my reply on the second outdated diff...

swertz:
@blinkseb I've added the `ingroups`.

PR #42, opened 2016-05-09
Basic info in README

swertz:
Add basic install instruction in the README. Anything else?

@blinkseb: I'm not sure about the minimal version & all requirements. Missing something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,57 @@
 # MoMEMta
-Modular Matrix Element Method implementation
+The Modular Matrix Element Method implementation.
+
+## Install
+
+### Prerequisites
+
+MoMEMta depends on the following libraries and tools:
+   * LHAPDF (>=6)
+   * CMake (>=2.8.8)
+   * Boost
+   * ROOT
+   * A C++11-capable compiler
+
+**Note**: MoMEMta has only been tested on GNU/Linux.
+
+### Procedure
+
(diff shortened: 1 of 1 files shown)


blinkseb:
Very nice thanks! Requirement seems ok. I'd remove `gcc` and state something like `a C++11 compatible compiler (tested with g++ 4.9 and 5.3)`

I'd also talk about the `-DCMAKE_INSTALL_PREFIX` option if you want to install `MoMEMta` elsewhere than the system directories

Something silly also, but I'd also add somewhere it's developed on `Linux` and was not tested on `OSX` nor `Windows`.

I'll make some comments on the text itself, but in any case, many thanks!

README.md
```diff
@@ -1,6 +1,55 @@
...
+
+### Prerequisites
+
+MoMEMta depends on the following libraries and tools:
+   * LHAPDF (>=6)
+   * CMake (>=2.8.8)
+   * Boost
+   * ROOT
+   * GCC (>=4.8)
+   * Lua (will be auto-installed if not available)
```
blinkseb: It's not really installed, we ship our own version with momemta but we don't install anything

README.md
```diff
@@ -1,6 +1,55 @@
...
+
+If you have admin rights on your system, you can make MoMEMta (public headers and library) available system-wide using:
+```
+# make install
+```
+
+### Build options
+
+The following options are available when configuring the build (when running `cmake ..`):
+   * `-DPROFILING=ON`: Generate debugging symbols
```
blinkseb: You also need to have `gperftools` installed somewhere if you turn-on `PROFILING` (https://github.com/gperftools/gperftools)

swertz:
Updated!

PR #45, opened 2016-05-11
Move initial partons from vector to pair

swertz:
To be inline with the matrix element exporter, move the intial parton object passed to the ME from a vector of 4-vectors to a pair of 4-vectors.


blinkseb:
Looks good thanks! Did you try to run it just to be sure?

swertz:
Yep!

blinkseb:
Thanks!

PR #46, opened 2016-05-13
Move to Boost.Log for logging

blinkseb:
This PR removes the current logger (spdlog) and replaces it by Boost.Log. The hard dependency we had with spdlog is now gone, since boost is more common and we already rely on it for other stuff.

Note: Travis has some issues currently, so I'm still not sure if this build or not on it, but it does build and work on my laptop.


core/src/logging.cc
```diff
@@ -16,25 +16,112 @@
...
-    namespace {
-        static std::shared_ptr<spdlog::logger> instance() {
-            bool in_terminal = isatty(fileno(stdout));
-            std::shared_ptr<spdlog::logger> logger = spdlog::stdout_logger_st("MoMEMta");
 
-            if (in_terminal)
-                logger->set_formatter(std::make_shared<colored_formatter>());
+typedef sinks::synchronous_sink<sinks::text_ostream_backend> text_sink;
+
+namespace {
```
swertz: Why is this empty namespace needed?

blinkseb: it's an anonymous namespace, it prevents access to the function from outside this file

swertz: Ok, did not know this trick!

swertz:
There is no build ongoing on travis... Expected?

blinkseb:
Yep :( See https://www.traviscistatus.com/ for more details

blinkseb:
Since this blocks basically everything else, I think it can be merged without waiting for travis. Worst case scenario, Travis build is broken, so it should not affect us too much...

swertz:
Tested: works both on @blinkseb's laptop and on ingrid -> we can merge.

PR #48, opened 2016-05-16
 Remove dependency to a private header in ParameterSet. Fixes #47

### LINKED ISSUE #47: Private header file included by public header
In https://github.com/MoMEMta/MoMEMta/blob/prototype/include/momemta/ParameterSet.h#L30 the file `lua/utils.h` is included, but this file is not included in the public headers.

Moving it to public headers would not solve the problem since the file includes `lua.hpp`: this would require the user to also specify the Lua include directory when linking something with MoMEMta, which I don't find desirable...

blinkseb:
There's also some fixes for Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ addons:
       - libboost1.55-dev
       - libboost-log1.55-dev
       - graphviz
+before_install:
   - ./travis/decrypt_key.sh 
 install:
   - export CXX=g++-4.9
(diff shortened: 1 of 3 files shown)


swertz:
Great, thanks!

swertz:
Well, I should have tested on ingrid before merging... I get the following:
```
CMake Error at CMakeLists.txt:168 (add_executable):
  Target "example_tt_fullyleptonic" links to target "Threads::Threads" but
  the target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?
```

swertz:
It's just a `find_package(Threads REQUIRED)` missing in the CMakeLists, I'll make a PR. Weird that it worked on travis??

PR #49, opened 2016-05-16
Find threads in CMake

swertz:
Fixes build on ingrid. @blinkseb, what was the reason for requiring pthreads actually?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ SET(CMAKE_C_FLAGS_DEBUG  "-O0 -g")
 # Find dependencies
 find_package(ROOT REQUIRED)
 find_package(LHAPDF 6.0 REQUIRED)
+find_package(Threads REQUIRED)
 
 if (NOT USE_BUILTIN_LUA)
     find_package(Lua 5.3 QUIET)


blinkseb:
That's weird, it should be automatically done by the `FindBoost` package. I guess the version of cmake on ingrid is too old?

Thanks for the fix :)

`pthread` is required for multiple reasons, but the main one is that Boost.Log depends on Boost.Thread which depend on `pthread` on unix. It's should be to ensure that the logger is thread-safe.

swertz:
Ok! I used the cmake 3.4 module, it's not that old...?

blinkseb:
I bumped travis version to 3.5.2 (the latest) and I test on my laptop with 3.5.1, so it's probably something they added in 3.5...

swertz:
Ok! I think we should not rely too much on the latest features, even if it's very tempting... ;) Some users might not have the choice of the version they use (even if a local installation of the latest is always an option).

blinkseb:
I totally agree. I already bumped min version to 3.2, because support for Boost in previous version is utterly broken and fixing it would be a real pain in the ass.

But unfortunately, they don't document from which release a particular features is available, so, unless browsing git history, you don't know if you are using something new or now (see this PR :smile: )

PR #50, opened 2016-05-17
[Docs] Use tables to describes modules' inputs and outputs

blinkseb:
As we discussed, this PR updates modules' documentation to use tables instead of lists to describe inputs and outputs.


swertz:
Thanks! I don't think we need to wait for travis to finish?

blinkseb:
Nop. I killed travis job, so you can merge right now

PR #53, opened 2016-05-17
Handle MET "more correctly"

swertz:
This is WIP, and any input is welcome.

- Change the `computeWeights` function to add the MET as optional parameter (defaults to a zero LorentzVector). The MET is then stored in the pool in the "input::met" input tag.
- BlockD, BlockB: Add the input "met" (if the user specifies nothing, "input::met" is assumed as default) and the optional parameter "pT_is_met": if at true, it fixes pT=-MET; if at false (default), it fixes pT=sum(visible), so the default behaviour corresponds to what we did before. When doing the sum(visible), it sums over **all** inputs, not just the 4 "blobs" of the block: the "inputs" vector can be longer than 4 entries to accomodate anything else the user wants to add (could be the output of a secondary block).
- BlockD: document
- BlockF: Also take all visible particles, but here the neutrinos **must** be reconstructed using p(tot)(vis), so the MET can not be used at all... (right?)
- Boost: renamed to "BuildInitialState"; define two modes depending on user's choice: do the transverse boost or not. Also, document.

To do:
- [x] Adapt other blocks
- [x] Useful Boost module
- [x] Document


modules/BlockD.cc
```diff
@@ -39,6 +40,15 @@ class BlockD: public Module {
             m_particle_tags = parameters.get<std::vector<InputTag>>("inputs");
             for (auto& t: m_particle_tags)
                 t.resolve(pool);
+
+            // If the met input is specified, get it, otherwise retrieve default
+            // one ("input::met")
+            if(parameters.exists("met")) {
```
blinkseb: Space between `if` and `(`

modules/BlockD.cc
```diff
@@ -55,13 +65,25 @@ class BlockD: public Module {
...
+            // We can either enforce momentum conservation by disregarding the MET, ie:
+            //  pT = sum of all the visible particles, 
+            // Or we can fix it using the MET given as input:
+            //  pT = -MET
+            // In the latter case, it is the user's job to ensure momentum conservation at
+            // the matrix element level (by using the Boost module, for instance).
+            
+            LorentzVector pT;
+            
+            if(pT_is_met) {
```
blinkseb: Space after `if`

modules/BlockD.cc
```diff
@@ -55,13 +65,25 @@ class BlockD: public Module {
...
+                pT = - m_met_tag.get<LorentzVector>(); 
+            } else {
+                LorentzVector ISR;
+                for(size_t i = 4; i < m_particle_tags.size(); i++) {
```
blinkseb: Space after `for`

blinkseb:
Many thanks for that, looks very good!

swertz:
Updated!

modules/BlockD.cc
```diff
@@ -22,14 +22,59 @@
...
+ * ### Integration dimension
+ *
+ * This module adds **0** dimension to the integration.
+ *
+ * ### Global parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `energy` | double | Collision energy. |
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$= \sum_{i \in \text{ vis} \vec{p}_i^{\text{vis}}\f$ |
```
blinkseb: This is not a `global` parameter: it's not read from `globalParameters()` but directly from the module parameters

modules/BlockB.cc
```diff
@@ -52,13 +52,15 @@
  *   | Name | Type | %Description |
  *   |------|------|--------------|
  *   | `energy` | double | Collision energy. |
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$= \sum_{i \in \text{ vis} \vec{p}_i^{\text{vis}}\f$ |
```
blinkseb: Same thing here, it's just a parameter

PR #56, opened 2016-05-23
Allow creation of non-existing parameters

blinkseb:
It's currently not possible to set or edit a parameter if it was not previously declared in the lua configuration file. For example, if the cuba verbosity is not defined in lua, it's not possible to change its value in C++.

With this PR, it's now possible to edit a parameter, even if it does not exist:
 - If the associated lua table does not exist, it's created first
 - The value of the parameter is changed directly in the lua side for `LazyParameterSet`, or simply in the map for `ParameterSet`

I've also simplified the `ParameterSet::set` method to perform implicit cast:
 - integral type to `int64_t`
 - floating point type to `double`

Previously, it was necessary to do explicit casting when doing for example 

```cpp
ParameterSet::set("value", (int64_t) 10);
```

There's also unit tests for the whole `ParameterSet` class.

--

See https://trello.com/c/Ej5PT07x for associated trello card

### DIFF
--- a/include/momemta/ParameterSet.h
+++ b/include/momemta/ParameterSet.h
@@ -23,6 +23,8 @@
 #include <memory>
 #include <string>
 
+#include <momemta/InputTag.h>
+#include <momemta/impl/traits.h>
 #include <momemta/Logging.h>
 
 #include <boost/any.hpp>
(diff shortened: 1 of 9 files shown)


include/momemta/ParameterSet.h
```diff
@@ -181,8 +208,20 @@ class ParameterSet {
...
+         * \param name Name of the new parameter
+         * \param value Value of the new parameter
+         */
+        virtual void create(const std::string& name, const boost::any& value);
         virtual void setInternal(const std::string& name, Element& element, const boost::any& value);
 
+        virtual void freeze();
+
+        std::map<std::string, Element> m_set;
+        bool frozen = false;
```
swertz: Why do these two become public?

swertz: Sry the 'protected' was hidden, my mistake!

blinkseb: They become `protected`, not `public`, because I need access to `m_set` from the derived class. `frozen` is actually not used, I've found another way. I'll re-put it `private`

swertz:
Awesome, thanks a lot!

blinkseb:
PR updated

PR #57, opened 2016-05-23
Push code coverage to coveralls.io

blinkseb:
Since we are starting to have some unit tests coverage with #56, I figured out it'd be nice to have some tracking and display on how many code lines are covered by unit tests. For that, there's the free service named [coverall.io](https://coveralls.io). Once merged, travis will start generating code coverage metrics, and push them to coveralls. We'll then be able to see the percentage of code covered by unit tests (spoiler alert, it's about 30% with #56, we can do better :)) and keep track across time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ addons:
       - graphviz
 before_install:
   - ./travis/decrypt_key.sh 
+  - pip install --user cpp-coveralls
 install:
   - export CXX=g++-4.9
   - export CC=gcc-4.9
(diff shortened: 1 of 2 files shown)


swertz:
Question: assuming we test that `computeWeights` returns always the same value, coverall will not know which modules we test by doing that since it doesn't know what is in the Lua file, right?

blinkseb:
I should be taken into account as well. Actual coverage is computed in travis when the executable is ran. IIUC, when compiling with the `--coverage` flags, some files are created indicating which lines contains executable code, and when an executable is ran, another file is created indicating which lines have been executed. Then, `gcov` is called and compute a nice summary using these files. The result is sent to coveralls, which only nicely display these data.

So if we decide to run some `computeWeights` tests, it should catch which modules are run, if the code for this module has been compiled with `--coverage` :)

swertz:
Ok, nice!

PR #60, opened 2016-05-24
Do not hardcode PDF renormalization scale

blinkseb:
Currently the PDF scale is hardcoded to the top mass. This PR adds a new mandatory input to the `MatrixElement` module for defining the renormalization scale used for the PDF.


swertz:
Thanks, that was needed indeed! @alesaggio you don't need to add it in your PR now... You'll want to rebase and adapt the examples...

I'll also have to update the tutorial.

PR #55, opened 2016-05-19
added WW fully leptonic

alesaggio:
Added WW_fullyleptonic.cc and WW_fullyleptonic.lua in MoMEMta/examples;
Added the matrix element from the exporter for WW fully lept in MoMEMta/MatrixElements;
MoMEMta/test not updated for this case yet;
Compiling and running

PR #62, opened 2016-05-26
Fix urgent bugs

swertz:
- Fix build of external ME (broken since #56)
- Correct check of physical solutions in blocks. @alesaggio I let you do it for blockF and include it in #55 

This show it'd be good to test the build of external ME when updating momemta...

Also, we need a review of the #include in all files: always include all what and only what is needed in each file.

### DIFF
--- a/core/src/InputTag.cc
+++ b/core/src/InputTag.cc
@@ -18,6 +18,7 @@
 
 
 #include <momemta/InputTag.h>
+#include <momemta/Utils.h>
 #include <momemta/Pool.h>
 
 std::vector<std::string> split(const std::string& s, const std::string& delimiters) {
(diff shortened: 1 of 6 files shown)


alesaggio:
What should I correct in the check of physical solutions? It seems working fine...

swertz:
Fix this https://github.com/MoMEMta/MoMEMta/blob/prototype/modules/BlockF.cc#L237 in a way similar to https://github.com/MoMEMta/MoMEMta/pull/62/files#diff-619955155ec5071ba66938630429679bR160
As we discussed yesterday (I think?): to check that the total energy/momentum is within physical bounds we have to consider all the inputs to the block.

PR #64, opened 2016-05-26
Explicitly write if a block is validated or not

blinkseb:
As discussed already. It'd be nice to also have this information somewhere else, maybe on the website on a `Validation` page?


modules/BlockD.cc
```diff
@@ -74,6 +74,8 @@
  *   |------|------|--------------|
  *   | `invisibles` | vector(vector(LorentzVector)) | LorentzVectors of the invisible particles, one entry (ie. one \f$(p_1, p_2)\f$ pair) per solution. |
  *   | `jacobians` | vector(double) | Jacobian of the performed change of variables (one entry per solution). |
+ *
+ * \note This module has been validated and is safe to use.
```
swertz: Maybe say "block" instead of "module"? It might lead to think that all the other modules are unsafe :p

blinkseb: Indeed, thanks :) PR Updated!

vidalm:
It has to be in the web somehow, Validation is fine. We can put there Block status and the link to MadWeight to see what a block means?

PR #65, opened 2016-05-27
[Docs] document MoMEMta class

blinkseb:
Fell free to suggest improvements and modifications!


swertz:
To me it's already a good start! Thanks a lot!

PR #67, opened 2016-05-31
[DOCS] Update readme; doc for remaining modules

swertz:
Small edit to the README, and add documentation for the remaining modules. Comments are welcome: documenting the matrix element module was especially tough...


modules/MatrixElement.cc
```diff
@@ -28,7 +28,125 @@
...
+ *
+ *    - 'visible' particles:
+ *      - `inputs` (vector(LorentzVector)): Set of visible particles.
+ *      - `ids`: Parameter set used to link the visibles to the matrix element (see below).
+ *
+ * ### Linking inputs and matrix element
+ *
+ * The matrix element expects the final-state particles' 4-momenta to be given in a certain order, 
+ * but it is agnostic as to how the particles are ordered in MoMEMta. 
+ * It is therefore necessary the specify the index of each input (visible and invisible)
```
blinkseb: **to** specify

blinkseb:
Thanks a lot, it's awesome. I found the matrix element documentation very well written, congrats! For a first iteration, it's more than enough, we'll see with time and feedback if some points need to be clarified! Thanks again!!

swertz:
Thanks for the feedback, PR updated.

I've cancelled the build since the code hasn't changed, and there'll probably be a second PR soon since we never manage to get all the Latex right the first time :p

blinkseb:
Thanks again!!!

PR #68, opened 2016-06-03
Move math functions from Utils.h to Math.h

blinkseb:
Move things from `Utils.h` to `Unused.h` and `Math.h`

 - `Math.h` contains all the maths functions
 - `Unused.h` contains the `UNUSED` declaration
 - `Utils.h` keeps functions I have no clue where to put otherwise :)

I also used `clang-formatter` to reformat the source code of math functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,7 @@ set(MOMEMTA_SOURCES
     "core/src/InputTag.cc"
     "core/src/LibraryManager.cc"
     "core/src/logging.cc"
+    "core/src/Math.cc"
     "core/src/MatrixElementFactory.cc"
     "core/src/MoMEMta.cc"
     "core/src/ModuleFactory.cc"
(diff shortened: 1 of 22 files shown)


swertz:
Thanks! This will break compatibility with the Tutorials: we need a way to follow the versioning of MoMEMta there... Maybe also parallel branches: a "v0.1.x" which is guaranteed to work with the "v0.1.x" of MoMEMta, and a "master" which might not always be up-to-date with the "master" or MoMEMta? And same thing with the exporter...

blinkseb:
Ah, you're right, forgot about that. Seems like a good strategy indeed

swertz:
Did you try to build external stuff (like a generated ME) with theses changes? The Tutorials are obviously broken; the ME should not be affected by this but since we've had issues with the include chains before I'd like to be sure...

blinkseb:
Nope, just did a full build of MoMEMta. But since I did not edit any matrix element files, risk of breakage are low. But indeed, it's better to check

swertz:
Works fine! Thanks again!

blinkseb:
Many thanks for testing!

PR #69, opened 2016-06-07
Improve handling of solutions with a Looper module

blinkseb:
Add looper module to loop over solutions

This is a major change in how the graph is evaluated. Previously, each module was responsible for looping over all possible solutions (called *invisibles*), and had to produce the same number of outputs at the number of inputs, otherwise the sync is lost. Dropping a solution was almost impossible.

Now, only the Blocks produce a set of solutions. All other modules expected in input only one solution. To deal with multiple solution, a new module is introduced, the Looper. As its name indicates, it loops over a set a solution and execute a set of modules for each iteration. In order for the Looper to know which modules it has to execute, a new concept is introduced, the Path. A Path is simply a collection of modules, specified by the user in the lua configuration file. Ordering is not important, since it also benefits from the graph feature we already have. Final module of a Path will usually be a `Summer` module, which simply sum some values at each iteration. To visualize the things, you will have a MatrixElement module inside a path, followed by a Summer module. The MatrixElement module will produce a weight for each solution, and the summer will sum all these weights to produce the final integrand.

It's possible that a solution provided by the Block is in reality not physical, but this can only be determined later on by another module. To solve the problem, the `Module::work` method now returns a value instead of `void`. Depending on the return value, the execution can be aborted, or simply skip to the next integration step. Possible values are:

 - `Status::OK`: everything is fine, keep going
 - `Status::NEXT`: Stop the current execution path, and skip to the next one. This has two different meaning:
    - If the module is in a loop, it means going to the next iteration of the loop, skipping remaining modules in the path
    - If the module is not in a loop, it means skipping the remaining modules, and going to the next integration step, returning 0 to cuba.
 - `Status::ABORT`: Desperate situation, abort the integration.

All modules have been updated to return `Status::NEXT` when they estimate that the current solution is non-physical.

As a module is now able to abort the integration, the user can retrieve the final status of the integration using the new `MoMEMta::getIntegrationStatus()` method, returning:

 - `IntegrationStatus::SUCCESS`: integration successful
 - `IntegrationStatus::ACCURARY_NOT_REACHED`: integration stopped before the desired level of accuracy was reached
 - `IntegrationStatus::FAILED`: integration failed
 - `IntegrationStatus::ABORTED`: integration aborted

To simplify the whole logic, a `Solution` class is added, wrapping together a set of LorentzVector and a jacobian. There's also a `valid` flag, not used for the moment.

Finally, there's one last change needed. Previously, MoMEMta tried to auto-detect which module produced the final value of the integrand, looking for a module producing an output named `integrands` and allowing only one of such module to exist. It's however no longer possible because it's now possible with the looper to have more than one module producing quantities that can be considered as final integrand. It's now mandatory to specify which InputTag to use to retrieve the integrand value, using the new `integrand()` lua function.

--

I've ported all the lua files to use the looper modules, and I've checked that both the TT & WW weight are the same before & after the change. I did not check the result of the phase-space volumes and cross-sections.

The documentation has also been updated to reflect the changes.

--

This is **not** in any way final, but I open the PR to gather feedback and comments on the feature. For example, I'm not really satisfied by the `Path` implementation. If you guys have better ideas!


vidalm:
Holy shit man, this is awesome.

vidalm:
I do not get why it was not possible to solutions before, and why the input and output had to exactly the same. (Let's talk in person). In any case the looper is much cleaner but I agree the path introduce an extra complication in the lua file... At some point the best documentation will be a lua example super commented.

vidalm:
Got it now, thanks!!

modules/Printer.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <momemta/Module.h>
+
+#include <sstream>
+
+#include <momemta/ParameterSet.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+
+/**
+ * \brief A module printing the value of an input
```
swertz: That's useful! Maybe add something like a "name" so that the user now what is being printed?

modules/DMEM.cc
```diff
@@ -55,10 +55,10 @@ class DMEM: public Module {
             m_hist->Reset();
         }
 
-        virtual void work() override {
+        virtual Status work() override {
             
             if (integrands->size() == 0)
```
swertz: This module would also have to be modified, not to do the loop itself (it would be included in the looper after the ME I guess)

blinkseb: Good catch!

modules/Printer.cc
blinkseb: Sure!

swertz:
Build is failing because of `c8149ec`...

blinkseb:
Yep, I forgot to update the unit tests :blush:

blinkseb:
Should work better now

swertz:
I've (finally) managed to read the PR... Big :+1: !! You said you we not completely happy with the implementation, but you've made changes since if I'm not mistaken? Is there still something you're not satisfied with?

Otherwise I must say I have no better idea, so no problem with merging (pending the comments to be fixed)...

blinkseb:
I have some ideas for refinement, but I'm afraid it'd take a while to implement, so it's probably better to go for this version first, and refine later when we have time.

If everyone is fine, I'll rebase the branch.

swertz:
For me it's good to go, once the DMEM module is modified ;)

blinkseb:
Thanks for the reminder :laughing:

blinkseb:
DMEM changes in ae0c4e5

blinkseb:
Hm the DMEM integral is not the same anymore. Do you see something wrong with my implementation?

swertz:
I think the DMEM module has to go in the looper: https://github.com/MoMEMta/MoMEMta/pull/69/commits/ae0c4e51f2949718346e304c27759e3047c564bd#diff-42ca085c71cbf316a382c47243b2b575R152

swertz:
Ha you've found it, the page didn't refresh...

Actually now I'm wondering if there shouldn't be a protection mecanism to warn the user when a module uses a solution but is not in the looper?

blinkseb:
Yep that's exactly what I thought too. Currently it's way to easy to mess things up (as I did actually :)). I'll see what I can do

blinkseb:
It was a bit harder than expected, but the execution is now stopped if a module uses the output of a looper without being in it's execution path.

Actually, a refinement at some point would be to build the path automatically using the list of modules dependent of the looper... To keep in mind!

blinkseb:
PR rebased

CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/) 
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [Unreleased]
+### Added
+ - New `Looper` module to loop over a set of solutions (see for more details)
```
swertz: see?

swertz:
To be sure I understand the new changes: you had to make some changes to be able to access the info here: https://github.com/MoMEMta/MoMEMta/pull/69/commits/009ab52de9d531c010638a14623659c82bab44aa#diff-57a27989d95956c8b94c36557dc4f602R181 and here: https://github.com/MoMEMta/MoMEMta/pull/69/commits/009ab52de9d531c010638a14623659c82bab44aa#diff-57a27989d95956c8b94c36557dc4f602R184, which means you needed access to the modules ParameterSet through the configuration, but because of lazy parameters you had to switch to pointers everywhere to make sure the same instances are accessed? Am I missing something?

blinkseb:
I needed to use pointer in order to avoid (again...) circular dependencies between `Configuration` and `ParameterSet`. And it's a mess because we want the ParameterSet to be copied when we freeze the configuration, otherwise modifying the configuration afterward would cause the changes to be propagate even to the freezed copy, etc.

swertz:
Ok, I see...

I find it increasingly confusing what we mean by 'configuration', 'module', 'lightModule', 'configuration::module', 'description' (I know it's hard to find self-explanatory names :D)... If people ever want to join the project it'd be good to have a bit of text describing the structure. Next time I'm bored I'll think about it...

Anyway, I'm good to go for the PR, thanks a lot again!

PR #71, opened 2016-07-11
normal_pdf issue corrected in GaussianTF

alesaggio:
Fixed the arguments passed to the function "normal_pdf".
dE_over_dP --> dP_over_dE


swertz:
Something's wrong. I'd advise you to create a new branch starting from the v0.1 branch, not the master, and cherry-pick your two commits there, then make a new PR.

blinkseb:
The cherry-pick won't work unfortunately for the dP_over_dE renaming, since the splitting of utils.h was not backported in v1. Only a new commit will work.

swertz:
Indeed. On the other hand, the dP_over_dE commit is not really a bugfix, so if you're lazy you can just backport the pdf thing (for which a cherry-pick will work)...

Issue #74, opened 2016-08-17
Two ways to access input tag content


swertz:
Right now we have two ways of accessing content of the memory pool by means of input tags:
1) Directly get a pointer to the content through `Pool::get<T>(tag)`, and dereference the pointer
2) Use the tag to access the content: the input tag has to be linked to the pool through `InputTag::resolve()`, and the content can then be retrieved through `InputTag::get<T>()`. This has to be used for indexed input tags, but can also be used in place of method 1).

A first point is that method 1) has less overhead than method 2), and should therefore be preferred whenever possible (for instance here: https://github.com/MoMEMta/MoMEMta/blob/master/modules/BlockD.cc#L95 one could have used method 1).

My worry is that method 1) forbids using indexed input tags, meaning that the user has less freedom when writing the config file. Although we haven't had a case showing that one would need the possibility to use either an indexed input tag or not, we should keep in mind that this might happen, and that we'll then have to enforce method 2) everywhere...

What do you guys think?

blinkseb:
I totally agree, we should unify the use of the pool by allowing only one entry point, supporting indexed and non-indexed `InputTag`.

My idea is to change the return value of `Pool::get<>` to return a `Value<T>` instead of a `std::shared_ptr<>`. `Value<T>` would have the same properties as `std::shared_ptr<T>` (`*` and `->` overloaded operators), but would act as a proxy between the user and the value. If the original `InputTag` is indexed, any wall to `*` or `->` would access the correct element in the array instead of the raw value. Oppositely, if the `InputTag` is non-indexed, the raw value would be cached, and any call would be directly transfered to the value.

I (for the moment) do not see any problem with this design, and it'd unify the two approach. What do you think?

swertz:
Seems like a good idea to me! Of course it adds some overhead compared to method 1), but if we want to unify there is no way around I'm afraid...

blinkseb:
Fixed by #76

PR #76, opened 2016-09-05
Return Value<T> from the Pool instead of plain shared_ptr

blinkseb:
As discussed in #74, I propose to change the return value of `Pool::get` from `shared_ptr` to `Value`. The new type acts as a proxy between the user and the value, and can automatically handle indexed and non-indexed input tags.

For performance reason, I did not include any check to ensure the index of the input tag is in the range of the vector.

As a bonus, we gain also the ability to declare a module using an indexed InputTag before the module producing the vector (it was not possible before), and some unit tests of the Pool and Value.

---

This PR is based on #69 . Relevant commits added in this PR can be seen here: https://github.com/blinkseb/MoMEMta/compare/95ce57e423b80c2bbbdef8bebd2eb11b11984926...9a9facebc88eeaf9725d3a782d0ce6f03882d0cb


swertz:
Thanks a lot, that was quick!

PR #80, opened 2016-09-06
Allow vector integrands

swertz:
Take advantage of Cuba's feature allowing vector integrands: each component is integrated in parallel, the grid being optimised for all of them (obviously, this only works if all components are similar).

Does not change much for the user: `MoMEMta::computeWeights()` is not modified. To define the components, simply give several input tags as arguments to the `integrand()` function in Lua:
```
integrand("integrand1::sum", "integrand2::sum")
```
integrates the output of modules `integrand1` and `integrand2` separately.

### DIFF
--- a/core/src/MoMEMta.cc
+++ b/core/src/MoMEMta.cc
@@ -66,14 +66,20 @@ MoMEMta::MoMEMta(const Configuration& configuration) {
 
     // Integrand
     // First, check if the user defined which integrand to use
-    InputTag integrand = configuration.getIntegrand();
-    if (integrand.empty()) {
+    std::vector<InputTag> integrands = configuration.getIntegrands();
+    if (!integrands.size()) {
         LOG(fatal) << "No integrand found. Define which module's output you want to use as the integrand using the lua `integrand` function.";
         throw integrands_output_error("No integrand found");
     }
 
+    // Next, retrieve all the input tags for the components of the integrand
     m_pool->current_module("momemta");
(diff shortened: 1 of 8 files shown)


core/src/MoMEMta.cc
```diff
@@ -137,11 +143,19 @@ std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector
     // Output from cuba
     long long int neval = 0;
     int nfail = 0;
-    double mcResult = 0, prob = 0, error = 0;
-
+    double* mcResult = new double[m_n_components]; 
+    double* prob = new double[m_n_components]; 
+    double* error = new double[m_n_components];
```
blinkseb: You have a leak here for the 3 arrays (no `delete[]` associated). Please use a `std::unique_ptr<double[]>` here

swertz:
Of course... Better like that?

swertz:
Darn...
```
Saving to: ‘ROOT-6.06.02_Python-2.7_gcc-4.9.3_Ubuntu_14.04_x86_64.tar.xz’

15% [=====>                                 ] 25,873,149  18.5KB/s  eta 99m 17s
```

blinkseb:
Can you possibly add a unit test for the `lua::set_final_module` function (in `tests/lua.cc`)?

swertz:
Done!

blinkseb:
Awesome thanks a lot!

PR #82, opened 2016-09-06
Change way PS points are retrieved & dimensions are counted

swertz:
Until now, each module would declare how many integration dimensions it needs via the `Module::dimension()` method, and the PS point input tags are declared in Lua either manually (`cuba::ps_points/i`), or using the `getpspoint()` function (which would create an input tag with the correct `i`).

This is not flexible enough to handle cases where a single PS point has to be used for several modules, without adding extra dimensions (typical case: evaluating a vector integrand, where each component of the integrand has to run on a same PS point, without changing the dimensionality of the integration phase space).

The way it now works is that `getpspoint()` is replaced by `add_dimension()`. It still creates and returns the correct input tag, but handles the counting of the dimensions itself via a callback which increases a counter in the `Configuration` class.

By doing for instance:
```
my_ps_point = add_dimension()
Module.module1 = { ps_point = my_ps_point }
Module.module2 = { ps_point = my_ps_point }
```
the `my_ps_point` can be used twice and the number of dimensions is still only 1.

The `Module::dimensions()` methods have been removed, the unit test and the examples updated.


blinkseb:
Very nice, thanks! Coverage even grows by 0.4%! :smile: 

 Can you also update the changelog to mention the change?

swertz:
Done!

blinkseb:
Perfect, thanks a lot!

PR #83, opened 2016-09-07
cmake and travis update

blinkseb:
Two things in this PR:
 - Since we require cmake 3.2, use modern cmake features like public and private headers / libraries. This cleans a bit the main cmake file. Matrix elements are also built automatically as soon as a new folder is added to the `MatrixElements` folder.
 - Travis now does some parallel build using different version of cmake (especially version 3.2.1 the minimum we require) and ROOT to ensure everything builds and works :)

And unrelated, change log #73


swertz:
Thanks! I don't know cmake well enough to comment, so since it still builds it's good enough ;)

PR #84, opened 2016-09-08
Start unit tests for modules

blinkseb:
I started to write tests for the modules, currently only for `BreitWignerGenerator` and `UniformGenerator` (yes, the easiest ones :laughing:). Once merged, I invite everyone to contribute and add a small test for a module. Simply testing that for a set of given input, the output is was you expected is already more than what we have currently. In the future, it'd good to think of a way to test more the internal of the module (specifically for the blocks) instead of just its response.


swertz:
Thanks a lot!
We should handle chocolate or beer to those who write tests for modules!

PR #86, opened 2016-09-09
Add unit test for blockD

swertz:
Tests for other blocks will be very similar.

The only thing is that it might be painful to find input LorentzVectors that actually give the desired number of solutions...

### DIFF
--- a/tests/unit_tests/modules.cc
+++ b/tests/unit_tests/modules.cc
@@ -29,6 +29,9 @@
 #include <momemta/Module.h>
 #include <momemta/ParameterSet.h>
 #include <momemta/Pool.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Math.h>
 
 #define N_PS_POINTS 5
 
@@ -38,6 +41,10 @@ public:
     ParameterSetMock(const std::string& name):
             ParameterSet(name, name) {
         // Empty
+    }
+
+    void createMock(const std::string& name, const boost::any& value) {
+        create(name, value);
     }
 };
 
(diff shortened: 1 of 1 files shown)


tests/unit_tests/modules.cc
```diff
@@ -147,4 +169,67 @@ TEST_CASE("Modules", "[modules]") {
...
+        *s25 = s_13_25;
+        *s256 = s_134_256;
+
+        std::vector<InputTag> inputs { 
+                { "input", "particles", 0 },
+                { "input", "particles", 1 },
+                { "input", "particles", 2 },
+                { "input", "particles", 3 }
+            };
+        parameters->createMock("inputs", inputs);
```
blinkseb: Is this needed because `set<std::vector<InputTag>>` does not work?

swertz: Yes exactly...

tests/unit_tests/modules.cc
```diff
@@ -147,4 +169,67 @@ TEST_CASE("Modules", "[modules]") {
...
\ No newline at end of file
+
+    SECTION("BlockD") {
+
+        parameters.reset(new ParameterSetMock("BlockD"));
+
+        double sqrt_s = 13000;
+        bool pT_is_met = false;
+
+        parameters->set("energy", sqrt_s);
```
blinkseb: I'm surprised this work since `energy` is part of the global parameters

blinkseb:
Awesome thanks a lot!! You earned your free beer for lunch!

tests/unit_tests/modules.cc
blinkseb: At some point we'll need support for that too if we want the user to be able to build the graph without configuration file, but I guess for the moment this will work :)

tests/unit_tests/modules.cc
swertz: I had noticed this issue, but indeed it works :)
It's thanks to this: https://github.com/MoMEMta/MoMEMta/blob/master/include/momemta/ParameterSet.h#L181

blinkseb: Nice :+1:

PR #87, opened 2016-09-09
Python bindings

blinkseb:
This PR introduce python bindings for MoMEMta, allowing to compute MEM weights directly from python :) Not all MoMEMta is exposed to python, but enough is to actually:
 - Parse configuration file
 - Edit parameters
 - Compute weights

Included in this PR is an integration test in python, computing weights and checking it's correct.

Bindings are not built by default, you have to add the `-DPYTHON_BINDINGS=ON` flag when running cmake. Python and boost::python are both needed for the bindings. If they are not available, bindings are not built. Also, if one want support for pyroot, then ROOT has obviously to be built with pyroot. If it's not the case, bindings are built, but without pyroot support.

TODO:
 - ~~Install target is missing for the python bindings~~ done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ before_script:
   - cd build
   - cmake .. -DTESTS=ON -DCOVERAGE=ON -DPYTHON_BINDINGS=ON -DCMAKE_BUILD_TYPE=Debug
 script:
-  - make -j2 && ./tests/unit_tests/unit_tests.exe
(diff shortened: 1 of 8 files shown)


swertz:
That's really awesome! :)

Can you update the changelog?

blinkseb:
Duh, of course :smile:

blinkseb:
Changelog updated!

swertz:
Muchas gracias!

PR #88, opened 2016-09-13
Add BlockA

alesaggio:


modules/BlockA.cc
```diff
@@ -0,0 +1,138 @@
...
+
+#include <momemta/Logging.h>
+#include <momemta/ParameterSet.h>
+#include <momemta/Module.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+#include <momemta/Math.h>
+
+/** \brief \f$\require{cancel}\f$ Final (main) Block A, describing \f$q_1 q_2 \to p_1 + p_2 + p_3 + p_4\f$
```
swertz: Please update the documentation to remove references to p3 and p4

modules/BlockA.cc
```diff
@@ -0,0 +1,138 @@
...
+            // pT = p1+p2+pb = 0. Equivalent to the following system:
+            // p1x+p2x = -pbx
+            // p1y+p2y = -pby,
+            // where: p1x=modp1*sin(theta1)*cos(phi1), p1y=modp1*sin(theta1)*sin(phi1),
+            //        p2x=modp2*sin(theta2)*cos(phi2), p2y=modp2*sin(theta2)*sin(phi2)
+            // Get modp1, modp2 as solutions of this system
+            // Note: Considering the bjets massless --> E = |p|
+
+            double modp1 = -pbx/(std::sin(theta1)*std::cos(phi1)) - (std::cos(phi2)/(std::sin(theta1)*std::sin(phi2-phi1)))*(pbx*std::tan(phi1)-pby);
+            double modp2 = (pbx*std::sin(phi1)-pby*std::cos(phi1))/(std::sin(theta2)*std::sin(phi2-phi1));
```
swertz: In principle you can use this function which solves a system of two linear equations: https://momemta.github.io/MoMEMta/dev/Math_8h.html#a4ae0b6426765b6a90a644a14247ac080

Plus, I'd like that quantities `std::sin(theta1)` be defined as (const) variables to avoid repeated calls to a complicated function...

modules/BlockA.cc
```diff
@@ -0,0 +1,138 @@
...
+ 
+            if(modp1 < 0 || modp2 < 0)
+               return Status::NEXT;
+
+            LorentzVector gen_p1(modp1*std::sin(theta1)*std::cos(phi1), modp1*std::sin(theta1)*std::sin(phi1), modp1*std::cos(theta1), modp1);
+            LorentzVector gen_p2(modp2*std::sin(theta2)*std::cos(phi2), modp2*std::sin(theta2)*std::sin(phi2), modp2*std::cos(theta2), modp2);
```
swertz: Here you shouldn't assume that p1 and p2 are massless.

modules/BlockA.cc
```diff
@@ -0,0 +1,138 @@
...
+
+            double inv_jac = (modp1*modp2)/(8*SQ(M_PI*sqrt_s));
+            inv_jac *= 1/std::abs(std::sin(phi2-phi1));
```
swertz: - Also you here you shouldn't assume that p1 and p2 are massless
- I think a part is missing in the denominator (in the MW reference I see also cos())
- The variable name is misleading, it is not an inverse jacobian but a jacobian

swertz:
Thanks a lot! :) There are a few things to fix...

At some point it'd be good to add a test as in #86, but you can do it later if you want ;)

modules/BlockA.cc
alesaggio: Ok, but here the problem is that the solve2Linear function requires a vector<double>, in my case I have just one solution (one pair (|p1|, |p2|)) (in principle I wouldn't need a vector) so I should always access the first element of the vector to retrieve them... is that ok anyway?

modules/BlockA.cc
alesaggio: The denominator is just written less explicitly (it's just the subtraction formula for the sine) ;)

swertz: Indeed :D But you still shouldn't simplify `|p|^2/E` to `|p|`...

modules/BlockA.cc
swertz: Yes that's true, it looks suboptimal...

But then you can check if there is really a unique solution (as we discussed, it might not always be the case) and handle the problematic case properly...

And it's better to re-use code whenever possible ;)

blinkseb:
In addition to Seb's comment, please also fix code styling. There should be a space between every math operation (so `a + b` and not `a+b`)

swertz:
And please also add a line to the changelog ;) https://github.com/MoMEMta/MoMEMta/blob/master/CHANGELOG.md

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+            const double sin_phi2 = std::sin(phi2);
+            const double cos_theta1 = std::cos(theta1);
+            const double cos_theta2 = std::cos(theta2);
+
+            const double a10 = sin_theta1*cos_phi1;
+            const double a01 = sin_theta2*cos_phi2;
+            const double a00 = -pbx;
+            const double b10 = sin_theta1*sin_phi1;
+            const double b01 = sin_theta2*sin_phi2;
+            const double b00 = -pby;
```
blinkseb: Missing some spaces here and above ;)

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+
+            bool foundSolution = solve2Linear(a10, a01, a00, b10, b01, b00, modp1, modp2, false);
+
+            if(!foundSolution)
```
blinkseb: Space after `if`

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+               return Status::NEXT;
+
+            if(modp1.size() > 1 || modp2.size() > 1)
```
blinkseb: Space after `if`

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+               return Status::NEXT;
+
+            double mod_p1 = modp1.at(0);
+            double mod_p2 = modp2.at(0);
+            if(mod_p1 < 0 || mod_p2 < 0)
```
blinkseb: Space after `if`

blinkseb:
Style is much better, thanks a  lot :)

modules/BlockA.cc
swertz: Here maybe it's not necessary to define all the aXX and bXX variables... But maybe it makes things more readable. As you wish!

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+#include <momemta/ParameterSet.h>
+#include <momemta/Module.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+#include <momemta/Math.h>
+
+/** \brief \f$\require{cancel}\f$ Final (main) Block A, describing \f$q_1 q_2 \to p_1 + p_2 + X\f$
+ *
+ * \f$q_1\f$ and \f$q_2\f$ are Bjorken fractions, and \f$p_1\f$, \f$p_2\f$, \f$p_3\f$ and \f$p_4\f$ are the 4-momenta of the visible
```
swertz: Also here remove reference to p3 and p4

blinkseb:
Can you also rebase to master please? Something like

```
git fetch origin
git rebase --onto origin/master HEAD^
```

should work

alesaggio:
Why can't I rebase this way? I get

CONFLICT (delete/modify): CHANGELOG.md deleted in HEAD and modified in add block A. Version add block A of CHANGELOG.md left in tree.

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+            LorentzVector gen_p1(mod_p1 * sin_theta1 * cos_phi1, mod_p1 * sin_theta1 * sin_phi1, mod_p1 * cos_theta1, E1);
+            LorentzVector gen_p2(mod_p2 * sin_theta2 * cos_phi2, mod_p2 * sin_theta2 * sin_phi2, mod_p2 * cos_theta2, E2);
+
+            double jacobian = (SQ(mod_p1) * SQ(mod_p2)) / (8 * SQ(M_PI*sqrt_s) * E1 * E2);
+            jacobian *= 1. / std::abs(std::sin(phi2 - phi1));
+
+            Solution s = { {gen_p1, gen_p2}, jacobian, true };
+            solutions->push_back(s);
+
+            return solutions->size() > 0 ? Status::OK : Status::NEXT;
```
swertz: At this point `solutions` will always have one entry (otherwise the function would have returned sooner), so you can simply return `Status::OK` without checking ;)

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+            const double b10 = sin_theta1 * sin_phi1;
+            const double b01 = sin_theta2 * sin_phi2;
+            const double b00 = -pby;
+
+            bool foundSolution = solve2Linear(a10, a01, a00, b10, b01, b00, modp1, modp2, false);
+
+            if (!foundSolution)
+               return Status::NEXT;
+
+            if (modp1.size() > 1 || modp2.size() > 1)
```
swertz: Why do you check that? I don't see why it would happen...

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+               return Status::NEXT;
+
+            double mod_p1 = modp1.at(0);
+            double mod_p2 = modp2.at(0);
```
swertz: It's nitpicking but while you're at it you can change `.at()` to `[]`. It's safe since we know the vector is not empty, and it's faster.

modules/BlockA.cc
```diff
@@ -0,0 +1,168 @@
...
+            if (mod_p1 < 0 || mod_p2 < 0)
+               return Status::NEXT;
+
+            double E1 = sqrt(SQ(mod_p1) + SQ(m1));
+            double E2 = sqrt(SQ(mod_p2) + SQ(m2));
+
+            LorentzVector gen_p1(mod_p1 * sin_theta1 * cos_phi1, mod_p1 * sin_theta1 * sin_phi1, mod_p1 * cos_theta1, E1);
+            LorentzVector gen_p2(mod_p2 * sin_theta2 * cos_phi2, mod_p2 * sin_theta2 * sin_phi2, mod_p2 * cos_theta2, E2);
+
+            double jacobian = (SQ(mod_p1) * SQ(mod_p2)) / (8 * SQ(M_PI*sqrt_s) * E1 * E2);
```
blinkseb: And finally, to be **really** annoying, you're missing a space between `M_PI` and `sqrt_s` :blush:

alesaggio:
Done ;)

modules/BlockA.cc
```diff
@@ -0,0 +1,165 @@
...
+               return Status::NEXT;
+
+            double E1 = sqrt(SQ(mod_p1) + SQ(m1));
+            double E2 = sqrt(SQ(mod_p2) + SQ(m2));
+
+            LorentzVector gen_p1(mod_p1 * sin_theta1 * cos_phi1, mod_p1 * sin_theta1 * sin_phi1, mod_p1 * cos_theta1, E1);
+            LorentzVector gen_p2(mod_p2 * sin_theta2 * cos_phi2, mod_p2 * sin_theta2 * sin_phi2, mod_p2 * cos_theta2, E2);
+
+            double jacobian = (SQ(mod_p1) * SQ(mod_p2)) / (8 * SQ(M_PI * sqrt_s) * E1 * E2);
+            jacobian *= 1. / std::abs(std::sin(phi2 - phi1));
```
blinkseb: Since here you already computed `sin / cos ( phi1 | phi2)` maybe it'd be interesting to replace `sin(phi2 - phi1)` by `sin(phi2)cos(phi1) - cos(phi2)sin(phi1)`? (see [here](https://fr.wikipedia.org/wiki/Identit%C3%A9_trigonom%C3%A9trique#Formules_d.27addition_et_de_diff.C3.A9rence))

alesaggio: Well yes... but isn't sin(phi2 - phi1) faster?

swertz: It would be, but as Seb said you aldready compute everything here: https://github.com/MoMEMta/MoMEMta/pull/88/files#diff-1c31af0a81a92b78fbd5601ae5714927R115 so you can re-use it: doing 2 products and 1 addition is faster than re-computing a sin()...

modules/BlockA.cc
```diff
@@ -0,0 +1,165 @@
...
+            const double sin_theta2 = std::sin(theta2);
+            const double cos_phi2 = std::cos(phi2);
+            const double sin_phi1 = std::sin(phi1);
+            const double sin_phi2 = std::sin(phi2);
+            const double cos_theta1 = std::cos(theta1);
+            const double cos_theta2 = std::cos(theta2);
+
+            const double a10 = sin_theta1 * cos_phi1;
+            const double a01 = sin_theta2 * cos_phi2;
+            const double a00 = -pbx;
```
swertz: I think the sign is wrong here (and below): the function solves a system where the constant term is as "ax + by + c = 0" (see documentation).

Even if we do not validate yet this block, could you please at least check on a dummy example that the solutions computed by the block satisfy the system of equations?

alesaggio: yes you're right, sorry about that, I'll do the test

swertz: Hm hang on, I might be the one who's wrong, I'll check that the function solve2Linear is correct first ^^

swertz: It's correct as it is!

swertz: *to be more clear, solve2Linear is correct

alesaggio: yes it's clear ;)

swertz:
Thanks again, sorry for all the iterations!

PR #89, opened 2016-09-14
T.F. "evaluators"

swertz:
- For the gaussian & binned TF on energy, add modules allowing to only evaluate the TF without actually integrating over an energy, which is necessary when using blocks such as BlockA or SecondaryBlockCD. This is done using base classes allowing to derive the TF itself and the "Evaluator" and maximise code re-use.
- It has been checked the value of the TF are consistent between the "actual" TF and the "evaluator" when using the same inputs.
- Rename "GaussianTransferFunction" to "GaussianTransferFunctionOnEnergy" for consistency
- For the Gaussian TF, the width of the Gaussian is now computed using the "gen-level" energy instead of the "reco-level" energy, which is more correct. However, when defining the range over which the integration is done, the "reco-level" energy is used (no choice there). This changes slightly the value of the weights from the examples.
- Documentation, changelog and examples updated
- The integration tests do not fail even if the value of the weights change -> to investigate


blinkseb:
Looks very nice, thanks a lot!

Coverage decreases because of added code in BinnedTransferFunction (not tested) and GaussianTransferFunction (evaluator is not tested).

:+1: for me!

PR #90, opened 2016-09-15
Edit matrix element parameters from the configuration file

blinkseb:
This PR adds a way to edit matrix element parameters from the configuration. This allow consistent parameter values for example between the BW generator and the matrix element. This is done by adding a new parameter to the Matrix Element module, `override_parameters`, which accept a `ParameterSet`: the key is the parameter name, and the value the new parameter value (a `double`).

Tested and works as expected. However, I'm opened to suggestion concerning the naming of the new parameter!


swertz:
Looks very good, thanks a lot!

I only have one worry: when the ME instance is built (here for instance: https://github.com/MoMEMta/MoMEMta/blob/master/MatrixElements/pp_ttx_fully_leptonic/SubProcesses/P1_Sigma_sm_gg_mupvmbmumvmxbx/cpp_pp_ttx_fullylept.cc#L136), some values of the parameters are used (to define the externel particles' masses) and will not be updated if we change them.

We might have to change the way this is handled in the exporter... But it's not too much of a worry since what is usually modified by the user during the computation are parameters such as internal masses, not external ones... What do you think?

examples/tt_fullyleptonic.cc
```diff
@@ -38,7 +38,7 @@ int main(int argc, char** argv) {
     ConfigurationReader configuration("../examples/tt_fullyleptonic.lua");
 
     // Change top mass
-    configuration.getGlobalParameters().set("top_mass", 173.);
+    // configuration.getGlobalParameters().set("top_mass", 173.);
```
swertz: I think this can be uncommented?

blinkseb: Sure

blinkseb:
This is tricky indeed. It'd be good if these masses in the ME were references to the original mass instead of copies. May be worth filling a bug ticket to not forget about it?

swertz:
Yes that should be enough!

PR #91, opened 2016-09-16
Change output of Looper module to a set of particles

blinkseb:
Currently, the Looper module output a `Solution`, which is supposed to be used by the various modules. But we can imagine some cases where two blocks produce solutions, which all must be used  to compute the matrix element. This is not possible for the moment.

This PR introduce a new behavior. The solution is gone, and the looper output only a set of particles, and the associated jacobian. This allows to drop the concept of visible and invisibles particles in most of the modules, and to use a more generic _particles_ concept. See the various configuration files to have an idea of the change needed.

 1. First commit is not directly related with this PR, but adds some context when a `boost::bad_any_cast` is thrown
 2. It's no longer possible to compute the phase space density for the final state particles inside the matrix element, because there's no distinction anymore between visible an invisibles particles. A new module is introduced, StandardPhaseSpace, which is responsible to compute this quantity. The user is then expected to pass this phase-space as a jacobian to the matrix element module.
 3. Second commit is a bug fix. In some cases, a module whose output is used by a module inside a looper can be scheduled *after* the looper instead of before. To avoid this, a fake connection is done between the looper and the module to force the execution *before* the looper.
 4. Last commit is more tricky. During development of this features, we observed some *extremely* weird behavior, with change in the final integrand of the examples. It has been traced to numerical instability in the computation of the final jacobian: `a * b * c` gives a different value than `a * c * b`. Difference is very small, but still present and enough to cause cuba to perform the integration differently. To mitigate this effect, we enabled the smoothing option of cuba by default.

While I was changing the modules, I optimized a bit the code for the matrix element module, like vectors that were filled for each step instead of once. I also tried to use reference to existing `LorentzVector` instead of doing copies when possible (see `BuildInitialState` for example)

I also edited a bit the module's documentation to reflect latest changes in the looper. Don't hesitate if you have some style comments, or things you want to see added!


blinkseb:
www.hepforge.org is down for the moment (for LHAPDF) so I killed travis jobs. Hopefully it'll come back tomorrow :)

modules/MatrixElement.cc
```diff
@@ -53,11 +52,13 @@
...
+ * \warning Keep in mind that the loop describe below is **not** done by this module. A Looper must be used for this.
+ *
  * If no invisibles are present and the matrix element only has to be evaluated on the uniquely defined visible particles,
  * this module can still be used to define the integrand: no loop is done in this case, and the index \f$j\f$ can be omitted
  * in the following.
  *
- * As stated above, this modules's ouput is a vector \f$\vec{I}\f$ whose entries are:
+ * As stated above, for each solution \f$j\f$, this modules's ouput is a scalar \f$I\f$:
  * \f[
  *      I_j = \frac{1}{2 x_1^j x_2^j s} \times \left( \sum_{i_1, i_2} \, f(i_1, x_1^j, Q_f^2) \, f(i_2, x_2^j, Q_f^2) \, \left| \mathcal{M}(i_1, i_2, j) \right|^2 \right) \times \left( \mathcal{J}^{\text{inv.}}_j \, \prod_{i} \mathcal{J}_i \right) \times \left( \prod_{i \in \text{vis.}} \frac{\left|p_i\right|^2 \sin(\theta_i)}{2 E_i(2\pi)^3} \right)
```
swertz: This formula needs to be updated to remove the final state phase-space, and remove the distinction between the block's and the other jacobians (same thing in the list below)

swertz:
Thanks a lot, looks very nice! There will be more work needed to update the website, but this can be done when we release...

+1 from my side, I don't see any obvious problems! The real test will be @alesaggio trying h(bb)h(bb) ;)

blinkseb:
Docs fixed, thanks! I also added a new test for the `ParameterSet::getNames` method

PR #92, opened 2016-09-19
Fix cuba Fork in release mode

blinkseb:
When building in release mode, `assert` is never executed, and so
the process is never forked.


swertz:
Thanks!

Question for backports: when we release v0.2, do we go on backporting to v0.1 or do we drop support?

blinkseb:
Good question! We basically know that 0.1 is not so useful given the addition we have, so I'm not in favor of supporting it at all once 0.2 is released. But we may need some kind of LTS at some point?

swertz:
Yes, but we're not there yet ;) That would v1.0...

swertz:
aaand we forgot the changelog :D

blinkseb:
:laughing:

PR #94, opened 2016-09-20
If n dimension is 0, do not perform integration

blinkseb:
If the configuration file does not request an integration dimension using `add_dimension`, there's no need for integration. If the number of dimension is 0, bypass cuba entirely, and evaluate the graph only once. In this case, no error is provided on the final result.


swertz:
Looks perfect, thanks a lot! Maybe output a LOG(debug) saying that no dimension was registered and that the integration will be bypassed?

Also, maybe you could add a line to the changelog for #92 ?

blinkseb:
PR updated :)

swertz:
Thanks again! Let's do some MELA...

PR #97, opened 2016-09-26
Handle nested loopers during validation. Fixes #95

### LINKED ISSUE #95: Problem with nested loopers
When nesting loopers, modules part of the execution path of the inside looper cannot access the output of modules which are part of the outer looper's path.

It might only be the Graph that doesn't understand what's happening, and that the execution itself will run fine...

blinkseb:
It's still not perfect, but at least does not throw when the configuration file is valid.

We need a better way to identify if a module actually depends on a looper or not..

### DIFF
--- a/core/src/Graph.cc
+++ b/core/src/Graph.cc
@@ -60,6 +60,22 @@ bool isConnectedTo(Graph& g, vertex_t vertex, vertex_t to) {
     return false;
 }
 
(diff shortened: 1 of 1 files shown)


swertz:
Thanks! It does solve part of the issue for nested loopers...

If I understand it right, however, this will not work if someones uses two loopers that are *not* nested, and a module inside one looper uses the other looper's output?

I'll merge this for now but we have to keep this in mind.

PR #98, opened 2016-09-26
Redirect cuba logging to our own logger

blinkseb:
Two benefits:
 - Output is prettier ( :laughing: )
 - One settings allow to shutdown all outputs from MoMEMta and cuba altogether.


swertz:
Thanks! Or course there is still Cuba's verbosity option, but we can keep that separated.

PR #101, opened 2016-09-29
Fix matrix element card parsing

swertz:
The piece of code retrieved from MadGraph supposed to parse the param cards doesn't work: for all parameters but those flagged as "decay" (ie widths), the reader would parse the line incorrectly, and the values in the card would not be used when instanciating a matrix element (default values are then used instead).

This should fix it. I've checked that I know get different values for the weight when changing ie. the top mass in the card.

It'll have to be backported to v0.1.

### DIFF
--- a/core/src/SLHAReader.cc
+++ b/core/src/SLHAReader.cc
@@ -1,10 +1,23 @@
 #include <algorithm>
 #include <fstream>
 #include <iostream>
+#include <regex>
 
 #include <momemta/SLHAReader.h>
 
 namespace SLHA {
+
+std::vector<std::string> parse_line_elements(const std::string& line, char delimiter=' ') {
+    std::vector<std::string> elements;
+    std::string temp;
+    std::stringstream linestr(line);
+    while (getline(linestr, temp, delimiter)) {
+        if (temp == std::string(&delimiter, 1))
+            continue;
+        elements.push_back(temp);
+    }
+    return elements;
+}
 
 class invalid_card_error: public std::runtime_error {
     using std::runtime_error::runtime_error;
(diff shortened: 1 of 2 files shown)


core/src/SLHAReader.cc
```diff
@@ -42,70 +55,82 @@ Reader::Reader(const std::string& file_name /* = ""*/) {
...
-            }
+    try {
+
+        while (std::getline(param_card, line)) {
+            // Change to lowercase
+            std::transform(line.begin(), line.end(), line.begin(), ::tolower);
+            // Remove comments
+            line = std::regex_replace(line, std::regex("#.*"), "");
+            // Remove leading, trailing and duplicate whitespaces
+            line = std::regex_replace(line, std::regex("^ *| 0*$"), "");
```
blinkseb: If I read this regex correctly, it removes any trailing `0` from the line, so `xxx 173000` will be truncated to `xxx 173`. Is it expected?

swertz: Good catch, that's actually a typo...

swertz:
PR updated. We might want to send this new version to the MG team at some point, since they use it for their default exporter...

blinkseb:
Thanks a lot for updating the changelog 😉 Concerning madgraph, do you think the C++ exporter is actually used? I had the feeling it was not. But still, we should definitely report this upstream. Thanks !

swertz:
I've reported it to MG but now I look like an idiot ^^ The bug was due to this change
https://github.com/MoMEMta/MoMEMta/commit/5f9a12c883b8daa172d62b36ac8bebb16ef37a1c#diff-6d62389ddb1f320f18961d6fbbcb3495R61
When defining the index as 'int' instead of 'double', a line such as "3 4.7" would be parsed as 3 (int), 4 (int), .7(double), instead of 3 (int) 4.7 (double) and the istream::operator>> would not fail when trying to read a "double-index block"...
Our new version works just as well, but the MG version was fine...

blinkseb:
... overlooked that during my refactoring, sorry about that!

swertz:
My bad, I also overlooked it ;)

PR #102, opened 2016-10-03
New functions for looper

### LINKED ISSUE #85: Typo in MoMEMta::IntegrationStatus enum
One field has a typo: `ACCURARY_NOT_REACHED` instead of `ACCURACY_NOT_REACHED`

https://github.com/MoMEMta/MoMEMta/blob/master/include/momemta/MoMEMta.h#L44

### LINKED ISSUE #96: Summer module does not reset its counter at the right moment
See https://github.com/MoMEMta/MoMEMta/blob/master/modules/Summer.cc#L35

Currently, the counter (or sum) is reset at the beginning of the Looper loop. This works correctly when only one Looper is used, but when nested, the sum is reset too soon.

The reset should happen in `beginIntegration` instead of `beginLoop`.

More over, it'd probably be good to rename the module to something like `LooperSummer`.

swertz:
Introduce two now functions for the modules: `beginPoint()` and `endPoint()`, called once for each module before the modules do their work, on each phase-space point.

This allows modules run in nested loops to initialise once at each PS point, ie fixes #96 

I've noticed that for modules inside a loop, functions such as `configure` or `beginIntegration` were not called. This fixes it.

Also, rename "Summer" to "LooperSummer" since it more accurately describes its use.

Unfortunately it means a lot of bogus calls at each PS point... If you have an idea to do it in a lighter way, that'd be welcome!

Also, take the opportunity to fix #85.


modules/Looper.cc
```diff
@@ -112,10 +112,30 @@ class Looper: public Module {
             path = parameters.get<Path>("path");
         };
 
+        virtual void configure() override {
+            CALL(configure);
```
blinkseb: The `configuration` function is already called for every modules here: https://github.com/swertz/MoMEMta/blob/281908f18e058943be463175e2e3bd713b6307d6/core/src/MoMEMta.cc#L57

swertz: Done!

PR #103, opened 2016-10-05
Add a way for the user to inject variables into the lua configuration

blinkseb:
As discussed yesterday, it's really useful to the user to inject variables into the lua configuration, for example to use transfer function or not, or to switch between muons or electrons, or whatever.

This is done by adding a new optional argument to `ConfigurationReader` constructor. It's a `ParameterSet`, and each entry of this set will be injected as a global variable into the lua state. You can see an example in the TTbar example, where I inject the `USE_PERM` and `USE_TF` variables.

First commit is not directly related to this PR, but decouples `ParameterSet` and lua, moving `LazyParameterSet` to the `lua/` subfolder (and now named `lua::LazyTable` to reflect what it really is)

cc @BrieucF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Main block A
  - Modules to only evaluate transfer functions (without integrating over anything): BinnedTransferFunctionOnEnergyEvaluator, GaussianTransferFunctionOnEnergyEvaluator
  - Matrix element parameters can now be edited from the configuration file.
+ - `ConfigurationReader` constructor now accepts an optional second argument allowing the definition of variables accessible from the lua configuration. 
 
 ### Changed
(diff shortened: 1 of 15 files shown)


swertz:
Looks good thanks a lot! Just a typo in the changelog...

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Main block A
  - Modules to only evaluate transfer functions (without integrating over anything): BinnedTransferFunctionOnEnergyEvaluator, GaussianTransferFunctionOnEnergyEvaluator
  - Matrix element parameters can now be edited from the configuration file.
+ - `ConfiguredReader` constructor now accepts an optional second argument allowing the definition of variables accessible from the lua configuration.
```
blinkseb: Indeed, fixed!

PR #104, opened 2016-10-05
Catch exception when creating a new module

blinkseb:
Display a log message telling which module was being constructed when the exception occurred. Can be useful for debugging.

Note: based on #103. I'll rebase once merged


examples/tt_fullyleptonic.lua
```diff
@@ -240,7 +240,7 @@ Looper.looper = {
       x_end = 2000.,
       n_bins = 500,
 
-      ps_weight = 'cuba::ps_weight',
+      ps_weight = 'cuba:ps_weight',
```
swertz: That shouldn't change ;)

blinkseb: :blush: indeed

blinkseb: And fixed

blinkseb:
Rebase :+1:

PR #106, opened 2016-10-06
Fix BlockA protection

BrieucF:


swertz:
Thanks, sorry for the trouble!

PR #107, opened 2016-10-06
CMSSW support

blinkseb:
This PR adds basic support for CMSSW:

  1. Cmake automatically detects the presence of a CMSSW env. and reacts accordingly by using LHAPDF and boost from the release instead of the system.
 2. When running install, MoMEMta is registered as a new tool inside the current release, which mean that any package can add MoMEMta as a dependency in their BuildFile.xml and use MoMEMta transparently

The main change here is that CMSSW does not ship Boost::log. In order to overcome this, I added a no-op logger used in case Boost::log is not present, which mean that there's *no* logging at all. A warning is issued during the cmake phase in this case.

I'm starting to think it may be good to drop boost::log for the logging as it's a bit of pain to manage, and maybe a bit overcomplicated for our needs...? But that's for another PR


swertz:
Thanks a lot, that's awesome!

Regarding boost log, it shouldn't be too hard to have our own logging system right? What was again the issue with spdlog that triggered to move to boost?

Can you update the changelog? :p Also, it'd be great if the options available when running cmake were all in the README (some are there but python and logging are missing)...

blinkseb:
At first, the idea was to use an already well-established library for logging that support multi-threading, filtering, logging redirection, etc. spdlog was good, but the issue was that it needs to be publicly available, since our public headers have some logging instructions. That means we need to install spdlog along with MoMEMta. Not really a huge problem, but it's still a lot of headers. That's what triggered the move to boost log, since it's actually "easier" to require boost log installed than spdlog.

Now, it's rather a pain to require such a big library for what we are doing with it, so I'd probably let me convince to move to a custom logging library... 😜

I'll update the changelog, as well as the readme 👍 . I'm also wondering wether the `NO_LOGGING` option is useful or not?

swertz:
Well, if we move away from Boost log, I think the `NO_LOGGING` option becomes superfluous since it's always possible to suppress the logging by changing the level...

I guess asking to make Boost log available in CMSSW would be too much of a pain?

blinkseb:
IMO it's near impossible. It tooks almost two months just to add scikit into CMSSW, and it's well motivated. I don't think they will allow a new lib just for MoMEMta.

I'll remove the `NO_LOGGING` option, and fix the rest :+1:

blinkseb:
Updated. I added support for Python also in addition to boost and LHAPDF.

swertz:
Thanks again!

PR #108, opened 2016-10-10
Add other integration algos

swertz:
Add Suave, Divonne and Cuhre as available in Cuba. To choose the algorithm, simply set the `algorithm` parameter in the Cuba global parameters in Lua (default to `vegas`).

The algos have lots of parameters that can be tweaked. I haven't tried to optimise them, so the default values are probably rubbish. Only experience will tell what is best (dedicace @BrieucF ). On the ttbar examples in MoMEMta, with these parameters Vegas converged always faster...

Divonne allows one to specify manually where the peaks are located. I've left it empty since for us it's almost impossible to know where the peaks are once the propagators have been removed. It might be possible in certain cases e.g. with transfer functions, so let's keep it in mind (not on the priority list for now)...


blinkseb:
Awesome, thanks a lot!

PR #109, opened 2016-10-11
Modify integrator default parameter

BrieucF:
Replace integrator default parameters by the values suggested [here](https://cubajl.readthedocs.io/en/latest/#cuhre-specific-keyword).


swertz:
Thanks!

PR #110, opened 2016-10-12
Fix transfer functions on energy

swertz:
Fix "bug" discovered by @BrieucF : transfer functions on energy now take as lower bound on Egen the mass of the given "reco" particle (lower bound was zero before, leading to NaNs if the object's mass was too high).


blinkseb:
Thanks!

PR #111, opened 2016-10-12
Move away from boost-log

blinkseb:
Boost-log is nice, but probably an overkill for our usage. Instead, I created a custom logger implementation, heavily inspired by spdlog. Multithreading is supported out of the box, could be useful in the near future. Added bonus, there's now logging in a CMSSW env :smile: 

It's almost 100% backward compatible. The only change is that level enums are now in `logging::level` and no longer in `boost::log::trivial`.


swertz:
Thanks a lot! We'll have to update the exporter to remove Boost log from the requirements (meaning the exporter will be incompatible with v0.1... we don't care?).

blinkseb:
I'd go with the _we don't care_ approach :smile:

PR #112, opened 2016-10-12
Sanity check

BrieucF:
The nan issue is fixed by #110

### DIFF
--- a/modules/BuildInitialState.cc
+++ b/modules/BuildInitialState.cc
@@ -41,6 +41,12 @@
  *
  * This module requires **0** phase-space point.
  *
+ * ### Global parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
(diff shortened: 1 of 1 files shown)


modules/BuildInitialState.cc
```diff
@@ -87,6 +95,10 @@ class BuildInitialState: public Module {
 
             do_compute_initials(particles);
 
+            // Check if solutions are physical
+            if (partons->at(0).E() > sqrt_s / 2 || partons->at(1).E() > sqrt_s / 2 )
```
blinkseb: No `at` please, it's slower than `[]`. Also it'd good to compute `sqrt_s / 2` once in the constructor instead of for each integration step

BrieucF: Updated

blinkseb:
Thanks!

PR #113, opened 2016-10-12
Add TFs on Pt

swertz:
It's a quick-and-dirty copy from the TFs on energy. If you think it's worthwile to find a better way to handle both possibilities to avoid duplicating a lot of code, I'll think about it, but it might make things less clear in the end...

Differences w.r.t. TF on energy:
- minimum value of Pt is 0, not the particle's mass (NOTE: some matrix elements might have soft divergences, so we might want to let the user set a minimum cut! same for energy)
- the way the "gen" 4-vector is defined is different
- the jacobian is different (d|P|/dPt = cosh(eta))


swertz:
Hang on, I'll change something...

swertz:
Updated. You can now specify an optional minimal Pt/energy to be considered when integrating on the TF. Useful for instance if the matrix element gamma->l+l- with massless leptons: you don't want to get too close to the divergence...

We might want to think about a better way of specifying cut on different quantities (using specific modules?)...

modules/BinnedTransferFunctionOnPt.cc
```diff
@@ -0,0 +1,269 @@
...
+class BinnedTransferFunctionOnPtBase: public Module {
+    public:
+
+        BinnedTransferFunctionOnPtBase(PoolPtr pool, const ParameterSet& parameters): Module(pool, parameters.getModuleName()) {
+            m_reco_input = get<LorentzVector>(parameters.get<InputTag>("reco_particle"));
+
+            m_file_path = parameters.get<std::string>("file");
+            m_th2_name = parameters.get<std::string>("th2_name");
+
+            m_file = std::unique_ptr<TFile>(TFile::Open(m_file_path.c_str()));
```
swertz: Because I've copied it from `BinnedTransferFunctionOnEnergy` where it's a class member :D

blinkseb:
Thanks a lot!

PR #114, opened 2016-10-13
Use internal copy of boost::any

blinkseb:
The goal here is to reduce as much as possible the need for external library when **using** MoMEMta. Using an internal copy of `boost::any` allows us to completely get rid of boost when using MoMEMta (note that boost is still needed when **building** MoMEMta)

This is 100% backward compatible. The only thing we need to change is the ME exporter to not look for boost.


swertz:
Thanks!

I think we're supposed to keep the Boost license somewhere, for instance in the preamble of the new `any.h` header...

blinkseb:
Done!

PR #115, opened 2016-10-13
Use cmake package files to ease finding MoMEMta

blinkseb:
cmake can automatically generate _package files_, which can be use by cmake itself to locate and configure packages: a `find_package(MoMEMta CONFIG REQUIRED)` is enough. This PR adds the necessary stuff to generate such files. Direct application of this is for the ME exporter: it's much easier to find MoMEMta this way (PR incoming)


swertz:
Very nice, thanks :)

PR #116, opened 2016-10-14
Fix build with GCC 6.2

blinkseb:
Ubuntu 16.10 is here, and it's shipping with gcc 6.2! This PR fix build with this version.


swertz:
Thanks! What's the change that makes this necessary?

blinkseb:
Not really sure, but:
 - A bunch of `M_PI`, `std::cos`, etc. not defined. I'm wondering how it was working before?
 - For cuba, I got a `# warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"`

PR #118, opened 2016-10-19
Validate main blocks A, B, G, secondary block C/D, fix blocks B and C/D

swertz:
- Add validation code (phase-space volume and cross-section) for main blocks A, B and G and secondary block C/D.
- Fix bugs in main block B and secondary block C/D that became apparent when running the validation
- All blocks can be considered validated (blockA cross section is still running... it's very slow! but the phase-space volume is good), the numbers are good within integration uncertainties (typically 0.5-1%)
- I'll put the numbers of the validation on the website along the others
- At some point I'll rewrite how the tests are done, to have only one executable for all tests instead of 10 copies of almost the same file...


swertz:
In principe the fix of block B should be backported to v0.1...

blinkseb:
Thanks a lot!

PR #120, opened 2016-10-19
Add DEBUG_TIMING cmake option to debug module runtime

blinkseb:
Can be useful to where most of the time is spent (spoiler, Matrix Element!). No impact if the flag `DEBUG_TIMING` is not set when running cmake.


swertz:
Very nice, thanks :)

PR #121, opened 2016-10-20
Clear paths before creating the graph

blinkseb:
Fix an issue where runtime was increasing each time a configuration reader was frozen. Modules inside each looper path were duplicated.


swertz:
Good that this was catched before the next release :D Thanks!

PR #123, opened 2016-10-21
Further fix to TF energy

BrieucF:
It looks like the check introduced in #110 was not sufficient to prevent `gen_pt` to be `nan`. Rejecting `gen_E < rec_M` is not enough either, only rejecting SQ(gen_E) < SQ(rec_M) did the trick... Looks like a precision issue. I don't know whether @swertz is ok with this way to fix it or not so I open the PR for discussion.

### DIFF
--- a/modules/BinnedTransferFunctionOnEnergy.cc
+++ b/modules/BinnedTransferFunctionOnEnergy.cc
@@ -155,7 +155,7 @@ class BinnedTransferFunctionOnEnergy: public BinnedTransferFunctionOnEnergyBase 
             const double delta = rec_E - gen_E;
 
             // To change the particle's energy without changing its direction and mass:
-            const double gen_pt = std::sqrt(SQ(gen_E) - SQ(rec_M)) / std::cosh(m_reco_input->Eta());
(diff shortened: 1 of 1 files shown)


swertz:
Wow, that's really strange... How come you did not have it at some point? What about rejecting `gen_E <= rec_M`?

BrieucF:
Not really sure why I did not had the problem before.
Nope, rejecting `gen_E <= rec_M` also leads to `nan gen_pt`...

swertz:
Sorry, can I ask you to try one more thing?

Here: https://github.com/MoMEMta/MoMEMta/pull/123/files#diff-682caa3d76971123340a79b521cd1fe1R160 instead of doing `SQ(gen_E) - SQ(rec_M)`, what happens if you do it instead as `(gen_E - rec_M) * (gen_E + rec_M)`?

blinkseb:
Looks like some numerical precision issue. In the case mentioned by Brieuc, we observed `gen_E - rec_M ~= e-5`

swertz:
As discussed IRL, this should be handled by #124, but if you'll add the missing square root to what it is at the moment I'll merge nonetheless ;)

BrieucF:
PR updated

PR #125, opened 2016-10-25
Ensure input p4 are physical

### LINKED ISSUE #124: Check inputs are physical
In https://github.com/MoMEMta/MoMEMta/blob/master/core/src/MoMEMta.cc#L122 we should check that inputs are physical (ie positive energy and mass), since otherwise it can lead to undefined behaviour and crashes because modules assume they are (and it would be a waste to check that in the modules).

We cannot possibly fix it ourselves so it's up to the user to give something correct as input.

blinkseb:
Fixes #124

### DIFF
--- a/core/src/MoMEMta.cc
+++ b/core/src/MoMEMta.cc
@@ -121,6 +121,11 @@ const Pool& MoMEMta::getPool() const {
 
 std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector<LorentzVector>& particles, const LorentzVector& met) {
 
+    for (const auto& p: particles)
+        checkIfPhysical(p);
+
+    checkIfPhysical(met);
+
     *m_particles = particles;
     *m_met = met;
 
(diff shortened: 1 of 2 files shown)


core/src/MoMEMta.cc
```diff
@@ -426,3 +431,25 @@ void MoMEMta::cuba_logging(const char* s) {
 MoMEMta::IntegrationStatus MoMEMta::getIntegrationStatus() const {
     return integration_status;
 }
+
+void MoMEMta::checkIfPhysical(const LorentzVector& p4) {
+    // Use M2() to prevent computation of the square root
+    if ((p4.M2() < 0) || (p4.E() <= 0)) {
```
swertz: I'd forgotten about the MET: since we set it as default to (0,0,0,0) the energy can be zero... That's why travis fails.

swertz:
Thanks!

core/src/MoMEMta.cc
```diff
@@ -121,6 +121,11 @@ const Pool& MoMEMta::getPool() const {
 
 std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector<LorentzVector>& particles, const LorentzVector& met) {
 
+    for (const auto& p: particles)
+        checkIfPhysical(p);
+
+    checkIfPhysical(met);
```
BrieucF: Are we sure we want to do this check also for met? It looks like we have a lot of met p4 with negative M2() in our DY n-tuples...

PR #127, opened 2016-11-02
Use named inputs in config files

blinkseb:
This is a change destinate to make configuration files more readable. Instead of presenting to the user inputs in a form `inputs::particles/X`, we now use named inputs. This works in two steps:

  1. On the configuration side, it's now mandatory to declare the inputs which are needed in the configuration, by using the `declare_input` function. One argument is needed, which is the name of the input. The function return an `input` object, which has the following attributes:
      - `name`: the name of the input
      - `reco_p4`: the input tag describing the `reco` part of the input, which is what the user gives to the `computeWeights` function
      - `gen_p4`: similar to `reco_p4`, but describing the `gen` part of the input. By default, it points to the same input tag as `reco_p4`, but it can be changed by the user by using the `set_gen_p4` function.

  2. On the C++ side, the `computeWeights` function now expects a vector of `Particle` instead of a vector of `LorentzVector`. A `Particle` is simply a small wrapper around a name, a type and a p4. Here, the names of the particles should match the names used in the configuration file. As a consequence, the ordering of the particles passed to the `computeWeights` function **does no longer matter**.

Suppose we have in the configuration:

```lua
local electron = declare_input("electron")
```
The user can use `electron.reco_p4` to access the input tag describing the input particle. This input tag will internally points to `electron::p4`. After a transfer function, he can set the gen input tag to the outut of the transfer function by using `electron.set_gen_p4("mytf::output")`, and then refer to this input tag using `electron.gen_p4`.

There's also some additions:

  - It's now possible to execute lua code before the configuration file is parsed. This code is located in the `lua/` folder. At build time, this code is embedded into the code by the `scripts/embedLua.py` script. This allows for example the functions `append` and `copy_and_append` to be built-in MoMEMta.
  - Two new lua functions are added, `add_reco_permutation` and `add_gen_permutations`, which will automatically add a permutator module permutating between the arguments of the function, and editing the reco and/or gen input tags to points to the output of the permutator (see for example the ttbar example). One has to be careful to call the function at the right place in the configuration file, as no check is done to ensure that the input tags have not been used before calling the function.
   
We also now needs to document the available lua functions somewhere. Maybe the doxygen documentation is the right place for that?


modules/BlockD.cc
```diff
@@ -104,7 +104,7 @@ class BlockD: public Module {
             if (parameters.exists("met")) {
                 met_tag = parameters.get<InputTag>("met");
             } else {
-                met_tag = InputTag({"input", "met"});
+                met_tag = InputTag({"met", "p4"});
```
BrieucF: Should'nt  we also change the description ```default `input::met``` at https://github.com/blinkseb/MoMEMta/blob/f611c3f5b4b488de39289c743ffeb285e1bd82c0/modules/BlockD.cc#L70 ?

examples/tt_fullyleptonic.cc
```diff
@@ -47,16 +47,16 @@ int main(int argc, char** argv) {
...
     // Muon
-    LorentzVector p5(-18.9018573760986, 10.0896110534668, -0.602926552295686, 21.4346446990967);
+    Particle muon { "muon", LorentzVector(-18.9018573760986, 10.0896110534668, -0.602926552295686, 21.4346446990967), +13 };
     // Anti b-quark
-    LorentzVector p6(71.3899612426758, 96.0094833374023, -77.2513122558594, 142.492813110352);
+    Particle b2 { "bjet2", LorentzVector(71.3899612426758, 96.0094833374023, -77.2513122558594, 142.492813110352), -5 };
 
     auto start_time = system_clock::now();
-    std::vector<std::pair<double, double>> weights = weight.computeWeights({p3, p4, p5, p6});
+    std::vector<std::pair<double, double>> weights = weight.computeWeights({muon, electron, b1, b2});
```
BrieucF: Some thoughts (not all related to this specific PR):
* It is a bit weird that the .cc associated to `tt_fullyleptonic.lua` does not pass met as input since the `pT_is_met` is set to true in this lua config. It would also provide an example on how to pass the MET (which is a bare p4 and not a Particle).
* IMHO the naming `pT_is_met` is misleading since the met is defined to be -PT of the system, if this flag is set to `false`. I would go for a more explicit `neutrino_pt_is_met`

modules/BlockD.cc
blinkseb: Indeed thanks!

examples/tt_fullyleptonic.cc
swertz: I agree on this point, but I don't like the `neutrino_pt_is_met` suggestion ;) The block makes no assumption that the reconstructed stuff is neutrinos...
If you don't mind you can leave it like that for now, I'll change the input tags anyway in a separate PR (as we discussed, to have the block inputs more explicit).

blinkseb: I agree, we should avoid explicit mention of a particle in the arguments :+1:

modules/BlockD.cc
blinkseb: Fixed!

blinkseb:
I fixed the documentation as requested by Brieuc, and I also added a bit of documentation about phase-space volume tests I have locally. It does not really belong here, but I'm too lazy to do a PR just for that :blush:

core/src/Graph.cc
```diff
@@ -249,7 +249,7 @@ Graph build(const Pool::DescriptionMap& description, std::vector<ModulePtr>& mod
     // Remove virtual vertices
     sorted_vertices.erase(std::remove_if(sorted_vertices.begin(), sorted_vertices.end(),
                 [&g](const vertex_t& vertex) {
-                    return Module::is_virtual_module(g[vertex].name);
+                    return Module::is_virtual_module(g[vertex].name) || !g[vertex].module;
```
swertz: Why was this needed?

core/src/MoMEMta.cc
```diff
@@ -129,14 +137,39 @@ const Pool& MoMEMta::getPool() const {
...
+            throw exception;
+        }
+
+        *m_inputs_p4[p.name] = p.p4;
+        *m_inputs_type[p.name] = p.type;
+
+        consumed_inputs.push_back(p.name);
+    }
 
     checkIfPhysical(met);
```
swertz: As discussed with @BrieucF, this can be removed

examples/WW_fullyleptonic.lua
```diff
@@ -1,39 +1,9 @@
...
-function copy_and_append(t1, t2)
-    local t3 = {}
-
-    append(t3, t1)
-    append(t3, t2)
-
-    return t3
-end
+-- Register inputs
+local electron = declare_input("electron")
```
swertz: Why necessarily `local`?

include/momemta/Particle.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <momemta/Types.h>
+
+#include <string>
+
+/**
+ * \brief Describe a reco particle. Used as input of MoMEMta::computeWeights
+ */
+struct Particle {
```
swertz: I'm wondering if we should start putting things into a `momemta` or `MoMEMta` namespace? `Particle` is a fairly common identifier...

lua/functions.lua
```diff
@@ -0,0 +1,142 @@
+-- Make an object immutable, preventing the user to modify it
```
swertz: This only prevents inserting new elements in a table, or am I missing something?

tests/phaseSpaceVolume/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Block B
+
+  - Runtime: 760.75user 0.98system 1:53.87elapsed 668%CPU
+  - 6.29735e-05 +- 5.8105e-08
+
+## Block D
+
+  - Runtime: 15481.15user 7.44system 38:41.82elapsed
+  - 645.063 +- 4.45919
```
swertz: This is really weird... We should try running it over a weekend, check again?

core/src/Graph.cc
blinkseb: `is_virtual_module` contains a list of hardcoded module we consider virtual, meaning it's not added by the user (`cuba` is one of them for example) and as a consequence the `module` pointer is null. Now, every input particle is a virtual module, but we cannot know that before the configuration file is parsed. One way is to change the `is_virtual_module` function and add a mechanism to update the list of virtual module by some way, or just checking if the `module` pointer is empty. The second one was the easiest :blush:

examples/WW_fullyleptonic.lua
blinkseb: It seems to be the preferred way to declare variable in lua, otherwise variables have global scope.

lua/functions.lua
blinkseb: No it prevents any modification of the object

include/momemta/Particle.h
blinkseb: And done :)

swertz:
Thanks again!

PR #128, opened 2016-11-23
SB_B, Vector Linear combination, no met check, fix nan

BrieucF:
This PR has four (some are unrelated) commits : 
* Implements the Secondary Block B usefull e.g. for single top processes
* Adds the possibility to perform four-vectors linear combination in the lua config
* Removes the check ensuring that the four-vector "met" is physical (at reco level, "met" has very often negative mass)
* Prevent nan values coming from numeric precision issues


modules/VectorLinearCombinator.cc
```diff
@@ -0,0 +1,87 @@
...
+ * ### Integration dimension
+ *
+ * This module requires **0** phase-space point.
+ *
+ * ### Inputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `inputs` | vector(LorentzVector) | List of LorentzVector used in the linear operation. |
+ *   | `coeficients` | vector(double) | List of the coeficients multiplying each LorentzVector in the linear combination (must have same size as `inputs`). |
```
blinkseb: `coefficient` and not `coeficient` :wink: every instance are misspelled :smiley:

modules/VectorLinearCombinator.cc
```diff
@@ -0,0 +1,87 @@
...
+            for (auto& v: vectors_tags)
+                m_vectors.push_back(get<LorentzVector>(v));
+
+            m_coeficients = parameters.get<std::vector<double>>("coeficients");
+        };
+
+        virtual Status work() override {
+
+            if (m_coeficients.size() != m_vectors.size()){
+                auto excepion = std::invalid_argument("VectorLinearCombinator recieved LorentzVector and Coeficient listis having different size.");
```
blinkseb: Can we have this sentence in English please :smile:

modules/VectorLinearCombinator.cc
```diff
@@ -0,0 +1,87 @@
...
+ * ### Outputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `output` | LorentzVector | LorentzVector resulting form the linear combination. |
+ *
+ * \ingroup modules
+ */
+
+class VectorLinearCombinator: public Module {
```
blinkseb: I think it'd be great to have this class templated, like it's done for the printer and stuff like that

modules/VectorLinearCombinator.cc
blinkseb: `exception` :wink:

BrieucF:
PR updated, can you check the templated Linear Cmbinator (did you had in mind something else than int double and 4-vectors)?

modules/LinearCombinator.cc
```diff
@@ -0,0 +1,91 @@
...
+            auto tags = parameters.get<std::vector<InputTag>>("inputs");
+            for (auto& v: tags)
+                m_terms.push_back(get<T>(v));
+
+            m_coefficients = parameters.get<std::vector<double>>("coefficients");
+        };
+
+        virtual Status work() override {
+
+            if (m_coefficients.size() != m_terms.size()){
```
blinkseb: Please move this check in the constructor.

modules/LinearCombinator.cc
```diff
@@ -0,0 +1,91 @@
...
+                auto exception = std::invalid_argument("The Term and Coefficient lists passed to LinearCombinator have different sizes.");
+                LOG(fatal) << exception.what();
+                throw exception;
+            }
+
+            T temp_result = m_coefficients[0] * *m_terms[0];
```
blinkseb: This will crash if the vector are empty. Please add a protection at the beginning of the `work` function

BrieucF:
PR updatedf

CMakeLists.txt
```diff
@@ -158,6 +158,7 @@ set(MOMEMTA_SOURCES
     "modules/BlockF.cc"
     "modules/BlockG.cc"
     "modules/SecondaryBlockCD.cc"
+    "modules/SecondaryBlockB.cc"
```
swertz: If we want to keep alphabetical order, this has to go one line up ;)

BrieucF:
Yes! And there is a typo :-)

modules/SecondaryBlockB.cc
```diff
@@ -0,0 +1,178 @@
...
+ *   | `s123` | double | Squared invariant mass of the propagator (GeV\f$^2\f$). |
+ *   | `p1` | LorentzVector | Temporary LorentzVector of the missing particle. It will be used only to retrieve the particle phi angle and mass.|
+ *   | `p2` | LorentzVector | Parton level LorentzVector of a decay product for which we already know everything.|
+ *   | `p3` | LorentzVector | Parton level LorentzVector of a decay product for which we already know everything.|
+ *
+ * ### Outputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `solutions` | vector(Solution) | Solutions of the equation \f$s_{12} = (p_1 + p_2)^2\f$. Each solution embed the parton level LorentzVector of the decay product with energy fixed by this equation and the jacobian associated with the change of variable mentioned above.|
```
swertz: Not only this equation, also the two other ones... Maybe not repeat it here, just reference what you explain above?

modules/SecondaryBlockB.cc
```diff
@@ -0,0 +1,178 @@
...
+            const double p3t = m_p3->Pt();
+            const double cosPhi12 = std::cos(ROOT::Math::VectorUtil::DeltaPhi(*m_p1, *m_p2));
+            const double cosPhi13 = std::cos(ROOT::Math::VectorUtil::DeltaPhi(*m_p1, *m_p3));
+            const double cosPhi23 = std::cos(ROOT::Math::VectorUtil::DeltaPhi(*m_p2, *m_p3));
+
+            const double p1t_linear = (E3 * p2z - E2 * p3z) / (cosPhi13 * p2z * p3t - cosPhi12 * p2t * p3z);
+            const double p1t_indep = (p2z * (2 * (E2 * E3 - p2z * p3z - cosPhi23 * p2t * p3t) + m3_squared - *s123 + *s12) - p3z * (m1_squared + m2_squared - *s12)) / (2. * (cosPhi13 * p2z * p3t - cosPhi12 * p2t * p3z));
+
+            const double p1z_linear = (cosPhi12 * E3 * p2t - cosPhi13 * E2 * p3t) / (-(cosPhi13 * p2z * p3t) + cosPhi12 * p2t * p3z);
+            const double p1z_indep = (-(cosPhi13 * p3t * (m1_squared + m2_squared - *s12)) + cosPhi12 * p2t * (2 * (E2 * E3 - cosPhi23 * p2t * p3t - p2z * p3z) + m3_squared + *s12 - *s123)) / (2 * (-cosPhi13 * p2z * p3t + cosPhi12 * p2t * p3z));
```
swertz: It looks like some things of the four lines above appear several times, with at most a different sign... Maybe those could be computed only once? Like e.g. the denominator `(cosPhi13 * p2z * p3t - cosPhi12 * p2t * p3z)`

modules/SecondaryBlockB.cc
```diff
@@ -0,0 +1,178 @@
...
+            bool foundSolution = solveQuadratic(E1_quadratic, E1_linear, E1_indep, E1_solutions);
+            if (!foundSolution)
+                return Status::NEXT;
+
+            // Use now the obtained E1 solutions to build p1
+            for (const double& E1: E1_solutions) {
+                // Skip unphysical solutions
+                const double p1t = p1t_linear * E1 + p1t_indep;
+                if (E1 <= m1 || p1t < 0)
+                    return Status::NEXT;
```
swertz: Here you shouldn't return but simply `continue` to the next `E1` solution

BrieucF: Nice catch :+1:

modules/SecondaryBlockB.cc
BrieucF: PR updated. I left the repetition of the equation to avoid saying twice mentioned above in the same sentence but if you really want, I can remove it ;-)

modules/SecondaryBlockB.cc
```diff
@@ -0,0 +1,187 @@
...
+            const double denominator = cosPhi13 * p2z * p3t - cosPhi12 * p2t * p3z;
+            const double E2E3 = E2 * E3;
+            const double p2zp3z = p2z * p3z;
+            const double p2tp3t = p2t * p3t;
+            const double E3p2z_E2p3z = E3 * p2z - E2 * p3z;
+            const double p1t_linear = E3p2z_E2p3z / denominator;
+            const double p1t_indep = (p2z * (2 * (E2E3 - p2zp3z - cosPhi23 * p2tp3t) + m3_squared - *s123 + *s12) - p3z * (m1_squared + m2_squared - *s12)) / (2. * denominator);
+
+            const double p1z_linear = (cosPhi12 * E3 * p2t - cosPhi13 * E2 * p3t) / - denominator;
+            const double p1z_indep = (-(cosPhi13 * p3t * (m1_squared + m2_squared - *s12)) + cosPhi12 * p2t * (2 * (E2E3 - cosPhi23 * p2tp3t - p2zp3z) + m3_squared + *s12 - *s123)) / (2 * - denominator);
```
BrieucF: Done ;-)

PR #129, opened 2016-11-28
Rename block inputs

swertz:
As we discussed: the inputs for the blocks are now names explicitly, and the (optional) further particles in the event (needed to balance the total transverse momentum, and check if solutions are physical) are put into the `branches` vector of input tags.

### DIFF
--- a/modules/BlockA.cc
+++ b/modules/BlockA.cc
@@ -88,7 +88,7 @@ class BlockA: public Module {
             auto branches_tags = parameters.get<std::vector<InputTag>>("branches");
             for (auto& t: branches_tags)
                 m_branches.push_back(get<LorentzVector>(t));
-            if (!m_branches.size()) {
+            if (m_branches.empty()) {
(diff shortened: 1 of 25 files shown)


modules/BlockA.cc
```diff
@@ -79,31 +82,36 @@ class BlockA: public Module {
...
-            auto particle_tags = parameters.get<std::vector<InputTag>>("inputs");
-            for (auto& t: particle_tags)
-                m_particles.push_back(get<LorentzVector>(t));
+            p1 = get<LorentzVector>(parameters.get<InputTag>("p1"));
+            p2 = get<LorentzVector>(parameters.get<InputTag>("p2"));
+            
+            auto branches_tags = parameters.get<std::vector<InputTag>>("branches");
+            for (auto& t: branches_tags)
+                m_branches.push_back(get<LorentzVector>(t));
+            if (!m_branches.size()) {
```
blinkseb: I'd prefer the use of `m_branches.empty()` instead of the negation

swertz: Done!

PR #131, opened 2017-01-28
Fix forwarding references and correct creation of std::shared_ptr using std::make_shared

Lezcano:
Changed  the details mentioned in #130.

PR #132, opened 2017-03-30
Update built-in lua to v5.3.4

blinkseb:
No new features, only a bug-fix release.


swertz:
Tanks! Maybe add a line in the changelog? (I realise we forgot to ask that for #131 )

blinkseb:
It's probably not useful to add a line for #131, we should only put impacting changes IMO. I'll add a line for the lua bump :+1:

swertz:
Ok, agreed. Thanks! Bypassing the build, it was successful...

PR #134, opened 2017-04-12
Add block E

swertz:
This PR adds the Main Block E. It's quite similar to block F but I've re-derived the equations allowing for non-zero masses for the invisible particles (PR for block F incoming).

There's no validation (yet) since we agreed not to add more executables; that will be for another PR.

### DIFF
--- a/modules/BlockE.cc
+++ b/modules/BlockE.cc
@@ -1,0 +1,272 @@
+/*
+ *  MoMEMta: a modular implementation of the Matrix Element Method
+ *  Copyright (C) 2017  Universite catholique de Louvain (UCL), Belgium
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <momemta/Logging.h>
+#include <momemta/Module.h>
+#include <momemta/ParameterSet.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Math.h>
+
+/*! \brief \f$\require{cancel}\f$ Final (main) Block E, describing \f$\hat{s} \to X + s_{13} (\to \cancel{p_1} p_3) + s_{24} (\to \cancel{p_2} p_4)\f$
+ *
+ * This Block addresses the change of variables needed to pass from the standard phase-space
+ * parametrisation to a parametrisation in terms of the invariant masses of the three propagators (\f$\hat{s}\f$, \f$s_{13}\f$ and \f$s_{24}\f$) and of the rapidity \f$y\f$ of the system.
+ * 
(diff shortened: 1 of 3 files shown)


modules/BlockE.cc
```diff
@@ -0,0 +1,269 @@
+/*
+ *  MoMEMta: a modular implementation of the Matrix Element Method
+ *  Copyright (C) 2016  Universite catholique de Louvain (UCL), Belgium
```
blinkseb: 2017 ;)

modules/BlockE.cc
```diff
@@ -0,0 +1,269 @@
...
+ */
+
+#include <momemta/Logging.h>
+#include <momemta/Module.h>
+#include <momemta/ParameterSet.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Math.h>
+
+#include <TMath.h>
```
blinkseb: Looks unused?

modules/BlockE.cc
```diff
@@ -0,0 +1,269 @@
...
+ * ### Integration dimension
+ *
+ * This module requires **2** phase-space points.
+ *
+ * ### Global parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|-------------|
+ *   | `energy` | double | Collision energy |
+ *   | `m1` <br /> `m2` | double, default 0 | Masses of the invisible particles \f$p_1\f$ and \f$p_2\f$ |
```
blinkseb: Those are not _global_ parameters, but parameters. Please add a new section `### Parameters`

modules/BlockE.cc
```diff
@@ -0,0 +1,269 @@
...
+             * Using the above, (1) is written as:
+             * a20 E2^2 + a02 p2y^2 + a11 E2 p2y + a10 E2 + a01 p2y + a00 = 0
+             *
+             * From (2)-(1), is it possible to write p2y = a E2 + b. This is then inserted into (1),
+             * yielding a quadratic equation in E2 only.
+             */
+
+            const double fac = - 2 * (B1x * pbx - B1z * ptotz - pby);
+            const double a = 2 * (A1x * pbx - A1z * ptotz - Etot) / fac;
+            const double b = (SQ(Etot) + pow(C1x + pbx, 2) + pow(C1z - ptotz, 2) + sq_m2 - SQ(C1x) - SQ(pby) - SQ(C1z) - sq_m1) / fac;
```
blinkseb: `SQ` can probably be used instead of `pow`?

modules/BlockE.cc
```diff
@@ -0,0 +1,269 @@
...
+                const double p2x = - p1x - pbx;
+                const double p2z = - p1z + ptotz;
+                const LorentzVector p2(p2x, p2y, p2z, E2);
+
+                // Check if solutions are physical
+                const LorentzVector tot = p1 + p2 + pb;
+                const double q1Pz = std::abs(tot.Pz() + tot.E()) / 2.;
+                const double q2Pz = std::abs(tot.Pz() - tot.E()) / 2.;
+                
+                if(q1Pz > sqrt_s/2 || q2Pz > sqrt_s/2)
```
blinkseb: Space after `if`

modules/BlockE.cc
swertz: I think this is a case where it would be counter-productive, since `SQ(a+b)` is replaced by `(a+b)*(a+b)` (two additions, one multiplication) whereas `pow(a+b,2)` does one addition, then one multiplication?

swertz:
Updated, thanks for the comments!

modules/BlockE.cc
blinkseb: Looks like indeed GCC is now smart enough to optimize `pow(x, 2)` to `x * x` so in the end both should give the same thing.

blinkseb:
Documentation looks fine, good to merge! Thanks a lot!

PR #135, opened 2017-04-12
Block F: support for massive invisibles

swertz:
As announced in #134, this PR adds support for massive invisible particles in block F.

Most of the code for the solution is actually the same between the two blocks... Not sure it's good to have so much duplicate code, but I don't see a clean way to avoid it while keeping things clear and efficient :/

Strange thing: when simply re-using the expressions of #134 for block F, the validation would converge much, much slower. The only difference was one equation solved in a slightly different way (mathematically equivalent), but it seems to have an important impact numerically. Since the way it was done previously in block F worked much better, I modified block E in this PR to do it in the same way.

Like that things seem to work but I confess I don't understand why one works better than the other...


vidalm:
Good to me, say hello to dark matter. I let the godfather (@blinkseb) comment.

blinkseb:
Just for confirmation, did you try to compute the phase-space volume and cross-section to ensure nothing broke? Otherwise it looks good to me too :+1: , thanks a lot!

swertz:
Yes I re-computed the PS volume with m=0 to check it was still OK. But we have no tests yet where we compute PS volumes with massive final state particles, that's something we could add at some point...

blinkseb:
Perfect, it's enough for the moment!

PR #136, opened 2017-04-13
Blocks B and D: support for massive invisibles

swertz:
NB: based on #135 

Modify Main Blocks B and D to support massive invisible particles.

I've also modified the secondary block A to harmonise the way the invisibles' masses are specified. For the moment the rule is the following:
- if the particle is fully reconstructed by the block, the mass is taken from a parameter
- if some quantity of the reconstructed particle is given by the user through an input 4-vector (typically, angles), the particle's mass is also taken from the input vector and not through a parameter (see e.g. the other secondary blocks).
Thoughts on this?

### DIFF
--- a/modules/BlockB.cc
+++ b/modules/BlockB.cc
@@ -40,7 +40,7 @@
  * - Conservation of momentum (with \f$\vec{p}_T^{tot}\f$ the total transverse momentum of visible particles):
  *  - \f$p_{1x} = - p_{Tx}^{tot}\f$
  *  - \f$p_{1y} = - p_{Ty}^{tot}\f$
- * - \f$p_1^2 = m_1^2 = 0\f$ (FIXME)
+ * - \f$p_1^2 = m_1^2\f$
  *
  * Up to two \f$p_1\f$ solutions are possible.
  *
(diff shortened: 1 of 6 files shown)


vidalm:
The parameter/input stuff make sense to me.

modules/BlockB.cc
```diff
@@ -172,7 +176,7 @@ class BlockB: public Module {
                 double q2Pz = std::abs(tot.Pz() - tot.E()) / 2.;
                 if(q1Pz > sqrt_s/2 || q2Pz > sqrt_s/2)
                     continue;
-
+
```
blinkseb: Spurious spacing

modules/BlockD.cc
```diff
@@ -247,7 +253,7 @@ class BlockD: public Module {
                         alpha6*e1 + beta6*e2 + gamma6,
                         alpha4*e1 + beta4*e2 + gamma4,
                         e2);
-
+
```
blinkseb: Spurious spacing

modules/BlockE.cc
```diff
@@ -208,20 +208,21 @@ class BlockE: public Module {
...
             
-            for (const double E2: E2_sol) {
+            for (const double p2y: p2y_sol) {
+                const double E2 = a * p2y + b;
+
                 if (E2 <= 0)
                     continue;
-
-                const double p2y = a * E2 + b;
+
```
blinkseb: And here too

swertz:
Fixed! No I did not recompute anything: when `m=0`, it amounts to add `0` to some variables, so it shouldn't change a thing :D

swertz:
Btw for block D there's a test so it would have picked it up: https://github.com/MoMEMta/MoMEMta/blob/master/tests/unit_tests/modules.cc#L173

blinkseb:
Ah great! Completely forgot about the unit test for Block D, we should definitively add ones for the other blocks!

PR #137, opened 2017-04-13
Add block C

fnand:
Hi,

So I have been attempting to write block C as specified in 1007:3300.

So far I have not been completely successful, and this should probably not be pulled into master, but into a new side branch. 

I am not sure whether my implementation is  faulty, or the values that I am using to test it result in ridiculous values (i.e. GIGO). 

I have added a new example as a test, as well as a new test in tests/phaseSpaceVolume/blockC

I have been working from blockD as a template. 
See https://github.com/fnand/MoMEMta/blob/add-blockC/BlockC2_eq.nb and/or https://github.com/fnand/MoMEMta/blob/add-blockC/BlockC.pdf for the algebra. 

It's possible a made a logic error somewhere along the way, or even a typo, but as far as I understand this is the way it should be done. Hopefully someone can spot my mistake. 

-Ferdi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Secondary blocks A, B, CD, E
  - Main blocks A, E and G (not present in MadWeight)
  - `getRandom4Vector` function to generate random Lorentz vectors of a specified mass (useful in cases where a particle has to be passed from C++, but integrated over all its components).
- - Main Block C and unit test for it
+ - Main Block C
 
 ### Changed
  - The way to handle multiple solutions coming from blocks has changed. A module is no longer responsible for looping over the solutions itself, this role is delegated to the `Looper` module. As a consequence, most of the module were rewritten to handle this change. See this [pull request](https://github.com/MoMEMta/MoMEMta/pull/69) and [this one](https://github.com/MoMEMta/MoMEMta/pull/91) for a more technical description, and this [documentation entry](http://momemta.github.io/) for more details
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Two new lua functions, `add_reco_permutations` and `add_gen_permutations` are available to easily insert a permutator module permutating between the function arguments. 
  - `LinearCombinator` templated module allowing to compute combinations (ie sums, subtractions, ...) of LorentzVectors, numbers, ...
  - Secondary blocks A, B, CD, E
- - Main blocks A, E and G (not present in MadWeight)
+ - Main blocks A, C, E and G (not present in MadWeight)
  - `getRandom4Vector` function to generate random Lorentz vectors of a specified mass (useful in cases where a particle has to be passed from C++, but integrated over all its components).
- - Main Block C
 
 ### Changed
  - The way to handle multiple solutions coming from blocks has changed. A module is no longer responsible for looping over the solutions itself, this role is delegated to the `Looper` module. As a consequence, most of the module were rewritten to handle this change. See this [pull request](https://github.com/MoMEMta/MoMEMta/pull/69) and [this one](https://github.com/MoMEMta/MoMEMta/pull/91) for a more technical description, and this [documentation entry](http://momemta.github.io/) for more details
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Secondary blocks A, B, CD, E
  - Main blocks A, E and G (not present in MadWeight)
  - `getRandom4Vector` function to generate random Lorentz vectors of a specified mass (useful in cases where a particle has to be passed from C++, but integrated over all its components).
+ - Main Block C and unit test for it
 
 ### Changed
  - The way to handle multiple solutions coming from blocks has changed. A module is no longer responsible for looping over the solutions itself, this role is delegated to the `Looper` module. As a consequence, most of the module were rewritten to handle this change. See this [pull request](https://github.com/MoMEMta/MoMEMta/pull/69) and [this one](https://github.com/MoMEMta/MoMEMta/pull/91) for a more technical description, and this [documentation entry](http://momemta.github.io/) for more details
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,7 @@ set(MOMEMTA_SOURCES
     "modules/BinnedTransferFunctionOnPt.cc"
     "modules/BlockA.cc"
     "modules/BlockB.cc"
+    "modules/BlockC.cc"
     "modules/BlockD.cc"
     "modules/BlockE.cc"
     "modules/BlockF.cc"
--- a/modules/BlockC.cc
+++ b/modules/BlockC.cc
@@ -167,12 +167,11 @@ class BlockC: public Module {
             const double beta2  = (sinthe3*sinphi3)/denom;
             const double gamma2 = -(2*E2*pTy - 2*X*pTy - *s12*sinthe3*sinphi3 + *s123*sinthe3*sinphi3)/denom;
 
-            const double alpha3 = 2*(-SQ(E2) + E2*X)/(-p2z*denom);
+            const double alpha3 = E2/p2z;
             const double beta3  = (p2x*cosphi3*sinthe3 + p2y*sinthe3*sinphi3)/(-p2z*denom);
-            const double gamma3 = ((-E2 + p2z*costhe3)*(SQ(m1) + p2Sq + 2*p2x*pTx + 2*p2y*pTy - *s12)
-                                   + 2*pTy*(SQ(p2x)*sinthe3*cosphi3 + SQ(p2y)*sinthe3*sinphi3)
-                                   + p2x*sinthe3*cosphi3*(SQ(m1) + p2Sq + 2*p2y*pTy - *s123)
-                                   + p2y*sinthe3*sinphi3*(SQ(m1) + p2Sq + 2*p2x*pTx - *s123))/(-p2z*denom);
+            const double gamma3 = 0.5*(-*s12 + SQ(m1) + p2Sq + 2*p2x*(pTx + sinthe3*cosphi3* 
+                                  (*s123 - *s12)/denom) + 2*p2y*(pTy + sinthe3*sinphi3 
+                                  *(*s123 - *s12)/denom))/p2z;
 
             const double beta4  = -1./denom;
             const double gamma4 = (*s123- *s12)/denom;
--- a/modules/BlockC.cc
+++ b/modules/BlockC.cc
@@ -65,7 +65,7 @@
  *   | `p2` <br/> `p3` | LorentzVector | The LorentzVector of \f$p2\f$ and the angles and mass of \f$p3\f$ will be used to reconstruct the event according to the above method. |
  *   | `branches` | vector(LorentzVector) | LorentzVectors of all the other particles in the event, taken into account when computing \f$\vec{p}_{T}^{branches}\f$ (if MET is not used), and checking if the solutions are physical. |
  *   | `met` | LorentzVector, default `met::p4` | LorentzVector of the MET |
- *
+ *   | `m1`  | Mass of the invisible particle, default 1 | 
  * ### Outputs
  *
  *   | Name | Type | %Description |
--- a/modules/BlockC.cc
+++ b/modules/BlockC.cc
@@ -55,7 +55,8 @@
  *
  *   | Name | Type | %Description |
  *   |------|------|--------------|
- *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$\vec{p}_{T}^{tot} = \sum_{i \in \text{ vis}} \vec{p}_i\f$ |
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$\vec{p}_{T}^{tot} = \sum_{i \in \text{ vis}} \vec{p}_i\f$ | 
+ *   | `m1`  | double, default 0 | Mass of the invisible particle | 
  *
  * ### Inputs
  *
@@ -65,7 +66,6 @@
  *   | `p2` <br/> `p3` | LorentzVector | The LorentzVector of \f$p2\f$ and the angles and mass of \f$p3\f$ will be used to reconstruct the event according to the above method. |
  *   | `branches` | vector(LorentzVector) | LorentzVectors of all the other particles in the event, taken into account when computing \f$\vec{p}_{T}^{branches}\f$ (if MET is not used), and checking if the solutions are physical. |
  *   | `met` | LorentzVector, default `met::p4` | LorentzVector of the MET |
- *   | `m1`  | Mass of the invisible particle, default 1 | 
  * ### Outputs
  *
  *   | Name | Type | %Description |
--- a/modules/BlockC.cc
+++ b/modules/BlockC.cc
@@ -1,0 +1,274 @@
+/*
+ *  MoMEMta: a modular implementation of the Matrix Element Method
+ *  Copyright (C) 2017  Universite catholique de Louvain (UCL), Belgium
+ *
+ *  This program is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include <momemta/ParameterSet.h>
+#include <momemta/Module.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Math.h>
+
+/** \brief \f$\require{cancel}\f$ Final (main) Block C, describing \f$s_{123} (\to p_3 + s_{12} (\to \cancel{p_1} p_2))\f$
+ *
+ * This Block addresses the change of variables needed to pass from the standard phase-space
+ * parametrisation for \f$p_{1 \dots 3} \times \delta^4\f$ to a parametrisation in terms of the two (squared) masses
+ * of the intermediate propagators and the angular variables of \f$p3\f$.
+ *
+ * The integration is performed over \f$s_{12}, s_{123}\, \theta_3, \phi_3\f$ with \f$p_{1 \dots 3}\f$ as input. Per integration point,
+ * the LorentzVector of the invisible particle, \f$p_1\f$, is computed as well as the energy of the (necessariy massless) particle \f$p3\f$, based on the following set
+ * of equations:
+ *
+ * - \f$s_{12} = (p_1 + p_2)^2\f$
+ * - \f$s_{123} = (p_1 + p_2 + p_3)^2\f$
+ * - Conservation of momentum (with \f$\vec{p}_T^{branches}\f$ the total transverse momentum of all branches represented):
+ *  - \f$p_{1x} + E_3 \sin\theta_3\cos\phi_3 = - p_{Tx}^{branches}\f$
+ *  - \f$p_{1y} + E_3 \sin\theta_3\cos\phi_3 = - p_{Ty}^{branches}\f$
+ *
+ * Up to four solutions are possible for \f$(E1, \alpha)\f$, where \f$\alpha = 2p_1\dotp_2\f$.
+ *
+ * ### Integration dimension
+ *
+ * This module requires **0** phase-space point.
+ *
+ * ### Global parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `energy` | double | Collision energy. |
+ *
+ * ### Parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$\vec{p}_{T}^{tot} = \sum_{i \in \text{ vis}} \vec{p}_i\f$ |
+ *
+ * ### Inputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `s12` <br/> `s123` <br/>  | double | Squared invariant masses of the propagators. Typically coming from a BreitWignerGenerator or NarrowWidthApproximation module. |
+ *   | `p2` <br/> `p3` | LorentzVector | The LorentzVector of \f$p2\f$ and the angles and mass of \f$p3\f$ will be used to reconstruct the event according to the above method. |
+ *   | `branches` | vector(LorentzVector) | LorentzVectors of all the other particles in the event, taken into account when computing \f$\vec{p}_{T}^{branches}\f$ (if MET is not used), and checking if the solutions are physical. |
+ *   | `met` | LorentzVector, default `met::p4` | LorentzVector of the MET |
+ *
+ * ### Outputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `solutions` | vector(Solution) | Solutions of the change of variable. Each solution embeds the LorentzVectors of the invisible particle (ie. \f$(p_1)\f$) and the massless particle (ie. \f$(p_3)\f$) and the associated jacobian. These solutions should be fed as input to the Looper module. |
+ *
+ * \note This block has NOT been validated and is NOT safe to use.
+ *
+ * \sa Looper module to loop over the solutions of this Block
+ *
+ * \ingroup modules
+ */
+
+class BlockC: public Module {
+    public:
+
+        BlockC(PoolPtr pool, const ParameterSet& parameters): Module(pool, parameters.getModuleName()) {
+            sqrt_s = parameters.globalParameters().get<double>("energy");
+            pT_is_met = parameters.get<bool>("pT_is_met", false);
+
+            s12  = get<double>(parameters.get<InputTag>("s12"));
+            s123 = get<double>(parameters.get<InputTag>("s123"));
+
+            m1 = parameters.get<double>("m1", 0.);
+            
+            p2 = get<LorentzVector>(parameters.get<InputTag>("p2"));
+            p3 = get<LorentzVector>(parameters.get<InputTag>("p3"));
+            
+            if (parameters.exists("branches")) {
+                auto branches_tags = parameters.get<std::vector<InputTag>>("branches");
+                for (auto& t: branches_tags)
+                    m_branches.push_back(get<LorentzVector>(t));
+            }
+
+            // If the met input is specified, get it, otherwise retrieve default
+            // one ("met::p4")
+            InputTag met_tag;
+            if (parameters.exists("met")) {
+                met_tag = parameters.get<InputTag>("met");
+            } else {
+                met_tag = InputTag({"met", "p4"});
+            }
+            m_met = get<LorentzVector>(met_tag);
+        };
+
+        virtual Status work() override {
+
+            solutions->clear();
+
+            // Don't spend time on unphysical corner of the phase-space 
+            if (*s12 >= *s123 || *s12 >= SQ(sqrt_s) || *s123 >= SQ(sqrt_s) )
+                return Status::NEXT;
+           
+            // pT will be used to fix the transverse momentum of the reconstructed neutrinos
+            // We can either enforce momentum conservation by disregarding the MET, ie:
+            //  pT = sum of all the visible particles,
+            // Or we can fix it using the MET given as input:
+            //  pT = -MET
+            // In the latter case, it is the user's job to ensure momentum conservation at
+            // the matrix element level (by using the Boost module, for instance).
+            LorentzVector pT;
+            if (pT_is_met) {
+                pT = - *m_met;
+            } else { 
+                pT = *p2;
+                for (size_t i = 0; i < m_branches.size(); i++) {
+                    pT += *m_branches[i];
+                }
+            }
+            
+            // p1x = alpha1 E1 + beta1 ALPHA + gamma1
+            // p1y = ...(2)
+            // p1z = ...(3)
+            // E3  = ...(4)            
+            const double cosphi3 = std::cos(p3->Phi());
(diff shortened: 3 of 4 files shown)


swertz:
Thanks a lot for this! You even included a unit test :+1:

Something I'd ask right away is for you to clean the pull request to keep only these three files:
- modules/BlockC.cc
- tests/unit_tests/modules.cc
- CMakeLists.txt

You can keep the rest for you in another branch of course. 
We're going to reorganise the tests soon (cross section & phase-space volume) so we'd like to avoid adding more stuff that's going to be changed anyway.
Of course if there's other things - not related to block C - that you would like to get included, you're welcome to open another PR!

Once that is done and the project compiles (for now the integration test fails) I'll pull your branch locally and check the block itself ;)

fnand:
Hi, 

I removed all the unnecessary changes, as well as the unit test I originally put in modules.cc

The test was basically just a skeleton and failed, meaning travis complained. My real test was the one in the phaseSpaces which I removed at your request.

swertz:
Great, thanks, I'll have a more in-depth look now :+1:.

Actually I think Travis failed because of some other reason... You're welcome to put the unit test back into `modules.cc`: from what I see the test itself is fine and should succeed if block C is correct!

fnand:
Hi, thanks, removed the latest commit then. 
The test is somewhat slapdash as it uses the input values from the other test (for blockD). 

The test without it failed with an error about the ttbar matrix element, which i am pretty sure I didn't mess around with.

tests/unit_tests/modules.cc
```diff
@@ -229,4 +229,53 @@ TEST_CASE("Modules", "[modules]") {
...
+
+        parameters->set("p2", InputTag("input", "particles", 0));
+        parameters->set("p3", InputTag("input", "particles", 1));
+
+        Value<SolutionCollection> solutions = pool->get<SolutionCollection>({"BlockC", "solutions"});
+
+        auto module = createModule("BlockC");
+
+        REQUIRE(module->work() == Module::Status::OK);
+        REQUIRE(solutions->size() == 2);
```
swertz: The rest of the test looks good; this is the only line that might have to be changed since there might be up to four solutions, and for the particular inputs used here the number of solutions has to be retrieved at least once from the block.

core/src/MoMEMta.cc
```diff
@@ -169,7 +169,7 @@ std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector
     }
 
     *m_met = met;
-
+
```
swertz: Please remove this

cmake/modules/FindLHAPDF.cmake
```diff
@@ -17,8 +17,8 @@ if (NOT LHAPDF_INCLUDE_DIR OR NOT LHAPDF_LIBRARY)
     find_program(LHAPDF_CONFIG_PROGRAM lhapdf-config)
 
     if (LHAPDF_CONFIG_PROGRAM)
-        exec_program(${LHAPDF_CONFIG_PROGRAM} ARGS "--incdir" OUTPUT_VARIABLE LHAPDF_INCLUDE_DIR)
-        exec_program(${LHAPDF_CONFIG_PROGRAM} ARGS "--libdir" OUTPUT_VARIABLE LHAPDF_LIBRARY_DIR)
+        exec_program(${LHAPDF_CONFIG_PROGRAM} ARGS "--incdir"  OUTPUT_VARIABLE LHAPDF_INCLUDE_DIR)
+        exec_program(${LHAPDF_CONFIG_PROGRAM} ARGS "--libdir"  OUTPUT_VARIABLE LHAPDF_LIBRARY_DIR)
```
swertz: Please revert these changes

CMakeLists.txt
```diff
@@ -297,6 +298,7 @@ if(EXAMPLES)
     set_target_properties(example_WW_fullyleptonic PROPERTIES OUTPUT_NAME
       "WW_fullyleptonic.exe")
 
+
```
swertz: Please remove the empty line

tests/unit_tests/modules.cc
```diff
@@ -229,4 +229,53 @@ TEST_CASE("Modules", "[modules]") {
...
+
+        for (const auto& solution: *solutions) {
+            REQUIRE(solution.valid == true);
+
+            LorentzVector test_p12  = input_particles->at(0) + solution.values.at(0);
+            LorentzVector test_p123 = input_particles->at(1) + test_p12;
```
swertz: Again, for `p3` you shouldn't use the input one but the block's output, ie `solution.values.at(1)` once you've added it.

swertz:
I think the expressions would get shorter and less error-prone if you defined `X=p2*p3/E3` since it appears in several places ;) Also computing it once and for all would make things quicker.

modules/BlockC.cc
```diff
@@ -0,0 +1,370 @@
...
+            //  pT = -MET
+            // In the latter case, it is the user's job to ensure momentum conservation at
+            // the matrix element level (by using the Boost module, for instance).
+
+            LorentzVector pT;
+            if (pT_is_met) {
+                pT = - *m_met;
+            } else {
+		//Err, do I need to add that other quark
+                pT = p2 + p3;
```
fnand: I'm not certain how this should work. How do the branches work?

fnand:
Hi, so I made some of the changes you suggested. I will more carefully look at tabs alignment soon, but my primary concern is with the fact that something doesn't work. 

One of the strange things is I often get messages similar to:

> No solution to equation 0 x^2 + 0 x + -5.96046e-08

other than that the s12 calculated is usually pretty far from the given value, which is an obvious clue that something is wrong. 

I will go through the equations again, but yes, CForm sucks and I eventually didn't use the CForm, just the normal output. 

Any idea what is wrong with my implementation, or how I can check it granularly?

fnand:
Some examples of outputs I get for calulated s12 vs given s12 is: 

`Calculated:  298761  Input   275480
Calculated:  201579  Input   247018
Calculated:  199799  Input   247018
Calculated:  212706  Input   233726
Calculated:  -266845  Input   318164
Calculated:  -266845  Input   318164
Calculated:  383650  Input   318164
Calculated:  357671  Input   245967
Calculated:  249490  Input   307394`

Which are significant differences

modules/BlockC.cc
swertz: So, the branches is all the other reconstructed/visible objects in the event, beside p1 and p3 which are the ones that get reconstructed by this block.
In this case this means only p2, plus the extra branches, which could be extra particles or could come from secondary blocks.

swertz:
I have found the problem :)

So the function `solve2Quads` doesn't work when `a11=0`, which is the case here. I'm sorry, this should be documented. You can simply solve it here: https://github.com/MoMEMta/MoMEMta/pull/137/files#diff-5d8ea00c7c37f24044bef758f6f78750R213 by renaming a<->b from lines 213-226.

Actually when defining the alphas (https://github.com/MoMEMta/MoMEMta/pull/137/files#diff-5d8ea00c7c37f24044bef758f6f78750R183 and so on), since some of them are zero, you can simply throw them away, and this will simplify the expressions for aXX and bXX...

modules/BlockC.cc
fnand: Great thanks!

swertz:
Ok I'll let you fix the remaining issues and the style, and then we'll have a final round ;)

fnand:
Perfect yeah that solves it. 

Will do some cleaning up and then commit later

fnand:
Done, see if this is ok?

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+        BlockC(PoolPtr pool, const ParameterSet& parameters): Module(pool, parameters.getModuleName()) {
+            sqrt_s = parameters.globalParameters().get<double>("energy");
+            pT_is_met = parameters.get<bool>("pT_is_met", false);
+
+            s12  = get<double>(parameters.get<InputTag>("s12"));
+            s123 = get<double>(parameters.get<InputTag>("s123"));
+
+            m1 = parameters.get<double>("m1", 0.);
+
+            m_particles.push_back(get<LorentzVector>(parameters.get<InputTag>("p2")));
```
swertz: Since there are only two we might avoid using a vector, which also avoids the step https://github.com/MoMEMta/MoMEMta/pull/137/files#diff-5d8ea00c7c37f24044bef758f6f78750R125? @blinkseb an opinion?

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+                auto branches_tags = parameters.get<std::vector<InputTag>>("branches");
+                for (auto& t: branches_tags)
+                    m_branches.push_back(get<LorentzVector>(t));
+            }
+
+            // If the met input is specified, get it, otherwise retrieve default
+            // one ("met::p4")
+            InputTag met_tag;
+            if (parameters.exists("met")) {
+
```
swertz: Please remove empty line

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+                for (size_t i = 0; i < m_branches.size(); i++) {
+                    pT += *m_branches[i];
+                }
+            }
+
+            // p1x = alpha1 E1 + beta1 ALPHA + gamma1
+            // p1y = ...(2)
+            // p1z = ...(3)
+            // E3  = ...(4)            
+            const double p2Sq   = p2.Dot(p2);
```
swertz: Please use `p2.M2()`

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+
+            const double pTx = pT.Px();
+            const double pTy = pT.Py();
+            
+            // Denominator that appears in several of the follwing eq.
+            // No need to compute it multiple times
+            const double denom = 2.*(E2 - p2z*costhe3 - p2x*cosphi3*sinthe3 - p2y*sinthe3*sinphi3 );
+                  
+            // Term appears regularly, compute once.    
+            const double X = p2x*sinthe3*cosphi3 + p2y*sinthe3*sinphi3 + p2z*costhe3;
```
swertz: Since it appears in `denom` it could be used there as well?

And actually we also have `X=p2.Vect().Dot(p3.Vect())/p3.P()` or `denom = 2*p2.Dot(p3)/p3.P()`... Not sure which method is quicker or more readable in the end, I'll leave it up to you.

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+
+            const double alpha3 = 2*(-SQ(E2) + E2*X)/(-p2z*denom);
+            const double beta3  = (p2x*cosphi3*sinthe3 + p2y*sinthe3*sinphi3)/(-p2z*denom);
+            const double gamma3 = ((-E2 + p2z*costhe3)*(SQ(m1) + p2Sq + 2*p2x*pTx + 2*p2y*pTy - *s12)
+                                   + 2*pTy*(SQ(p2x)*sinthe3*cosphi3 + SQ(p2y)*sinthe3*sinphi3)
+                                   + p2x*sinthe3*cosphi3*(SQ(m1) + p2Sq + 2*p2y*pTy - *s123)
+                                   + p2y*sinthe3*sinphi3*(SQ(m1) + p2Sq + 2*p2x*pTx - *s123))/(-p2z*denom);
+
+            const double beta4  = -1./denom;
+            const double gamma4 = ( *s123- *s12)/denom;
```
swertz: Please leave no space after an opening parenthesis and before a closing parenthesis (also in the lines below).

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+
+            // Find the intersection of the 2 conics (at most 4 real solutions for (E1,ALPHA))
+            std::vector<double> E1, ALPHA;
+            solve2Quads(a11, a22, a12, a10, a01, a00, b11, b22, b12, b10, b01, b00, E1, ALPHA, false);
+
+            // For each solution (E1,ALPHA), find the neutrino 4-momentum p1
+            if (E1.size() == 0)
+                return Status::NEXT;
+
+            for(unsigned int i=0; i<E1.size(); i++){
```
swertz: A few spaces missing -> `for (std::size_t i = 0; i < E1.size(); i++) {`

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+
+                //Make sure E1 is not negative
+                if (e1 < 0.) 
+                    continue;
+            
+                const double e3 = beta4*alp + gamma4;   
+                // Make sure E3 is not negative 
+                if (e3  < 0.) 
+                    continue;
+
```
swertz: Please only one empty line (also below)

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+                    tot += *m_branches[i];
+                }
+                double q1Pz = std::abs(tot.Pz() + tot.E()) / 2.;
+                double q2Pz = std::abs(tot.Pz() - tot.E()) / 2.;
+                if(q1Pz > sqrt_s/2 || q2Pz > sqrt_s/2)
+                    continue;
+
+                double jacobian = computeJacobian(p1, p2, p3);
+                
+                Solution s {{p1, p3}, jacobian, true };
```
swertz: Either add a space after the first `{` or remove the one after the last `}` ;)

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+                solutions->push_back(s);
+            }
+
+            return solutions->size() > 0 ? Status::OK : Status::NEXT;
+        }
+
+        double computeJacobian(const LorentzVector& p1, const LorentzVector& p2, const LorentzVector& p3) {
```
swertz: In my previous round of comment I suggested to move all this part in the loop above, to avoid re-accessing all the vector components that have already been computed, ie. define `p1x = beta1*alp + gamma1`, and so on, and use them both to define `p1` and `p2` and compute the jacobian.

tests/unit_tests/modules.cc
```diff
@@ -229,4 +229,53 @@ TEST_CASE("Modules", "[modules]") {
...
+            REQUIRE(solution.valid == true);
+
+            LorentzVector test_p12  = input_particles->at(0) + solution.values.at(0);
+            LorentzVector test_p123 = solution.values.at(1) + test_p12;
+            
+            REQUIRE(test_p12.M2()  == Approx(s_12));
+            REQUIRE(test_p123.M2() == Approx(s_123));
+
+ //         LorentzVector test_pT = test_p134 + test_p256;
+ //         REQUIRE(test_pT.Pt() == Approx(0));
```
swertz: You can uncomment (and adapt) the Pt() test ;)

You can also require that the input and output masses of `p1` and `p3` are "equal".

tests/unit_tests/modules.cc
```diff
@@ -229,4 +229,53 @@ TEST_CASE("Modules", "[modules]") {
...
+        }
+    }
+
+
+
```
swertz: Please max. one empty line

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
+/*
+ *  MoMEMta: a modular implementation of the Matrix Element Method
+ *  Copyright (C) 2016  Universite catholique de Louvain (UCL), Belgium
```
swertz: 2017

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ */
+
+
+#include <momemta/ParameterSet.h>
+#include <momemta/Module.h>
+#include <momemta/Solution.h>
+#include <momemta/Types.h>
+#include <momemta/Math.h>
+
+/** \brief \f$\require{cancel}\f$ Final (main) Block C, describing \f$q_1 q_2 \to \f$X + s_{123} (\to p_3 + s_{12} (\to \cancel{p_1} p_2))\f$
```
swertz: Please remove `q_1 q_2 \to \f$X + ` as they are not present in the final parameterisation

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * This Block addresses the change of variables needed to pass from the standard phase-space
+ * parametrisation for \f$p_{1 \dots 3} \times \delta^4\f$ to a parametrisation in terms of the two (squared) masses
+ * of the intermediate propagators.
```
swertz: ... and the angular variables of `p3`

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * The integration is performed over \f$s_{12}, s_{123}\f$ with \f$p_{1 \dots 3}\f$ as input. Per integration point,
```
swertz: same here, please mention `theta_3` and `phi_3`

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ * the LorentzVector of the invisible particle, \f$p_1\f$, is computed based on the following set
```
swertz: ... and also `p3`'s energy
Maybe that's the also right place to mention that `p3` is required to be massless.

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$\vec{p}_{T}^{tot} = \sum_{i \in \text{ vis}} \vec{p}_i\f$ |
+ *
+ * ### Inputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `s12` <br/> `s123` <br/>  | double | Squared invariant masses of the propagators. Typically coming from a BreitWignerGenerator or NarrowWidthApproximation module. |
+ *   | `p1` ... `p3` | LorentzVector | LorentzVectors of the particles used to reconstruct the event according to the above method. |
```
swertz: `p2` is not required as input, so you can replace the dots by `<br/>`. Also please mention that for `p3` only angles and mass are used, as in https://github.com/MoMEMta/MoMEMta/blob/master/modules/BlockA.cc#L62

modules/BlockC.cc
```diff
@@ -0,0 +1,303 @@
...
+ *   | `branches` | vector(LorentzVector) | LorentzVectors of all the other particles in the event, taken into account when computing \f$\vec{p}_{T}^{branches}\f$ (if MET is not used), and checking if the solutions are physical. |
+ *   | `met` | LorentzVector, default `met::p4` | LorentzVector of the MET |
+ *
+ * ### Outputs
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `solutions` | vector(Solution) | Solutions of the change of variable. Each solution embed  the LorentzVectors of the invisible particle (ie. \f$(p_1)\f$) and the associated jacobian. These solutions should be fed as input to the Looper module. |
```
swertz: `Each solution embeds`
Also, `p3` should be mentioned here

modules/BlockC.cc
blinkseb: Indeed please avoid the `vector` here, no need to be dynamic since you know in advance you only have 2 particles. Just declare `p2` and `p3` directly as member of the class.

modules/BlockC.cc
blinkseb: Actually from the code inputs are `p2` and `p3`.

swertz: You're right, that was a typo ;)

modules/BlockC.cc
fnand: I switched around `X` and `denom` and used `X` in `denom`. 
I did not change it to the vector implementation, do you think it would be faster that way? It has a div in there so I am not sure, but I haven't tested it.

modules/BlockC.cc
fnand: Done

modules/BlockC.cc
fnand: Done

modules/BlockC.cc
fnand: Makes sense. Done.

modules/BlockC.cc
fnand: I guess if I was really concerned with optimization I could calculate 1/denom and then multiply by it...

modules/BlockC.cc
fnand: Done. But would you prefer `std::size_t`? All the other blocks use `unsigned int`. It's probably just a matter of consistency in this case as we don't expect large amounts of solutions.

modules/BlockC.cc
fnand: Done

modules/BlockC.cc
fnand: I must have missed that one. The reason I did it this way is because this is how it is done in BlockD. 

Maybe I am misunderstanding the logic, but wouldn't moving the return statement exit the loop prematurely?

fnand: Ah, you mean move the computeJacobian into the loop?

fnand:
Hi, 

Please check if the documentation is ok now. I wrote it before changing anything and subsequently forgot to update it. 

Also, the unit test is not yet complete. 

Right now I am just hijacking the values that were specified for block D. 
What would the suggested way to handle the inputs for BlockC be?

modules/BlockC.cc
swertz: Yes exactly as in there ;) https://github.com/MoMEMta/MoMEMta/blob/master/modules/BlockE.cc#L248

modules/BlockC.cc
```diff
@@ -0,0 +1,296 @@
...
+#include <momemta/Types.h>
+#include <momemta/Math.h>
+
+/** \brief \f$\require{cancel}\f$ Final (main) Block C, describing \f$s_{123} (\to p_3 + s_{12} (\to \cancel{p_1} p_2))\f$
+ *
+ * This Block addresses the change of variables needed to pass from the standard phase-space
+ * parametrisation for \f$p_{1 \dots 3} \times \delta^4\f$ to a parametrisation in terms of the two (squared) masses
+ * of the intermediate propagators and the angular variables of \f$p3\f$.
+ *
+ * The integration is performed over \f$s_{12}, s_{123}\, \theta_3, \phi_3f$ with \f$p_{1 \dots 3}\f$ as input. Per integration point,
```
swertz: There's a `\` missing before the `$`

modules/BlockC.cc
```diff
@@ -243,9 +236,9 @@ class BlockC: public Module {
                 if(q1Pz > sqrt_s/2 || q2Pz > sqrt_s/2)
                     continue;
 
-                double jacobian = computeJacobian(p1, p2, p3);
-                
-                Solution s {{p1, p3}, jacobian, true };
+                double jacobian = computeJacobian(p1, *p2, p3);
```
swertz: I would call the `p3` defined L223 differently, e.g. `p3_sol` or something like that. It works because you're in the loop, but it's not the same `p3` as above so it's not ideal for readability...

swertz:
Thanks! For the unit test, it's perfectly OK to re-use the same inputs as for block D! I think the test just fails because `p3` has to be massless for block C, and the "random" 4-vectors we're using to test block D are massive. So what you can do is add an extra "random" massless 4-vector, in the same way as it's done for the others, and use that one as `p3` input to test block C ;)

fnand:
I think all the issues except the unit test have been addressed now. Anything that I missed?

modules/BlockC.cc
```diff
@@ -0,0 +1,274 @@
...
+                LorentzVector tot = pT + p1 + p3_sol;
+                for (size_t i = 0; i < m_branches.size(); i++) {
+                    tot += *m_branches[i];
+                }
+                double q1Pz = std::abs(tot.Pz() + tot.E()) / 2.;
+                double q2Pz = std::abs(tot.Pz() - tot.E()) / 2.;
+                if(q1Pz > sqrt_s/2 || q2Pz > sqrt_s/2)
+                    continue;
+
+                const double p1x = p1.Px();
```
swertz: Instead of retrieving those here, you can define `p1x = beta1*alp + gamma1`, ..., and then use those both to define `p1` and `p3_sol` and to compute the jacobian.

swertz:
Thanks! I only had one small comment. For the unit test there's really not much missing ;)

blinkseb:
Thanks a lot for your contribution! Before we can merge it, you'll need to rebase your PR in top of `master`, and also squash your changes into a single commit, maybe two if you want to keep the unit tests separated from the main block. It'd be also nice if you can add a line to the change log, stating the block C was added!

Thanks again! :+1:

swertz:
Btw if you rebase on master you'll be able to use one of https://github.com/MoMEMta/MoMEMta/blob/master/tests/unit_tests/modules.cc#L81 as input for `p3` to test block C ;)

fnand:
Hi, 

So rebase has been completed. Commits have all been squashed into one.

Everything is still not fine with the unit test. Which either means the values I am giving it are unreasonable, or something is fishy in block c. 

I get   6394.8875476811 == Approx( 6400.0 ) currently. 

Also, is it just me or do I have to use input 0, otherwise the module won't accept it?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Secondary blocks A, B, CD, E
  - Main blocks A, E and G (not present in MadWeight)
  - `getRandom4Vector` function to generate random Lorentz vectors of a specified mass (useful in cases where a particle has to be passed from C++, but integrated over all its components).
+ - Main Block C and unit test for it
```
blinkseb: Please just edit line 25 to add `C` (no need for the unit test in the change log)

modules/BlockC.cc
```diff
@@ -0,0 +1,274 @@
...
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `energy` | double | Collision energy. |
+ *
+ * ### Parameters
+ *
+ *   | Name | Type | %Description |
+ *   |------|------|--------------|
+ *   | `pT_is_met` | bool, default false | Fix \f$\vec{p}_{T}^{tot} = -\vec{\cancel{E_T}}\f$ or \f$\vec{p}_{T}^{tot} = \sum_{i \in \text{ vis}} \vec{p}_i\f$ |
```
blinkseb: I see that your block accepts `m1` as a parameter but it's not documented. Can you add a line?

fnand: Done

CHANGELOG.md
fnand: Done

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
  - Secondary blocks A, B, CD, E
  - Main blocks A, E and G (not present in MadWeight)
  - `getRandom4Vector` function to generate random Lorentz vectors of a specified mass (useful in cases where a particle has to be passed from C++, but integrated over all its components).
- - Main Block C and unit test for it
+ - Main Block C
```
blinkseb: I meant to remove this line altogether and edit line 25 to read `Main blocks A, C, E and G (not present in MadWeight)`

modules/BlockC.cc
```diff
@@ -65,7 +65,7 @@
  *   | `p2` <br/> `p3` | LorentzVector | The LorentzVector of \f$p2\f$ and the angles and mass of \f$p3\f$ will be used to reconstruct the event according to the above method. |
  *   | `branches` | vector(LorentzVector) | LorentzVectors of all the other particles in the event, taken into account when computing \f$\vec{p}_{T}^{branches}\f$ (if MET is not used), and checking if the solutions are physical. |
  *   | `met` | LorentzVector, default `met::p4` | LorentzVector of the MET |
- *
+ *   | `m1`  | Mass of the invisible particle, default 1 |
```
blinkseb: Please respect the structure of the table, there're 3 column, `name`, `type` and `description`. Default mass is 0, not 1.

```
| `m1` | double, default to 0 | Mass of the invisible particle | 
```

And it's not an input, but a parameter, please move the line to the correct section of the documentation.

fnand: Ah, great, and I also wrote the incorrect default value. Fixing it now

swertz:
Hi Ferdinand,

After checking extensively on our side, it does look like something's broken again with the block's solutions...

If I remember well, it did work perfectly at some point, so I'm guessing one of the changes you've implemented has broken it, unless you did check regularly that the solutions were correct (which would have been a good thing to do)? 

Unfortunately we can't access either your commit history or the mathematica files to check where the problem might be...

Would you mind checking again the different expressions? I think it has to be somewhere  between here https://github.com/MoMEMta/MoMEMta/pull/137/files#diff-5d8ea00c7c37f24044bef758f6f78750R157 and here https://github.com/MoMEMta/MoMEMta/pull/137/files#diff-5d8ea00c7c37f24044bef758f6f78750R178, the rest looks fine to me.

modules/BlockC.cc
```diff
@@ -0,0 +1,274 @@
...
+            // No need to compute it multiple times
+            const double denom = 2.*(E2 - X);
+                  
+            const double beta1  = (cosphi3*sinthe3)/denom;
+            const double gamma1 = -(2*E2*pTx - 2*X*pTx - *s12*cosphi3*sinthe3 + *s123*cosphi3*sinthe3)/denom;
+
+            const double beta2  = (sinthe3*sinphi3)/denom;
+            const double gamma2 = -(2*E2*pTy - 2*X*pTy - *s12*sinthe3*sinphi3 + *s123*sinthe3*sinphi3)/denom;
+
+            const double alpha3 = 2*(-SQ(E2) + E2*X)/(-p2z*denom);
```
swertz: `alpha3 = E2 / p2z`

modules/BlockC.cc
```diff
@@ -0,0 +1,274 @@
...
+            const double beta3  = (p2x*cosphi3*sinthe3 + p2y*sinthe3*sinphi3)/(-p2z*denom);
+            const double gamma3 = ((-E2 + p2z*costhe3)*(SQ(m1) + p2Sq + 2*p2x*pTx + 2*p2y*pTy - *s12)
```
swertz: `gamma3 = 0.5 * (- *s12 + SQ(m1) + p2Sq + 2 * p2x * (pTx + sinthe3 * cosphi3 * (*s123 - *s12) / denom) + 2 * p2y * (pTy + sinthe3 * sinphi3 * (*s123 - *s12) / denom) ) / p2z`

swertz:
Found the problem, see the two comments.

Now the only thing that's missing in the unit test is to require that the input and output masses for `p1` are equal, and that `p3.M / p3.E = 0`.

modules/BlockC.cc
fnand: This solved it. I am trying to check where I went wrong. I suspect a typo in copying from mathematica.

fnand:
Thanks a lot. I am trying to figure out what went wrong. I suspect an error when copying from mathematica. I added the check for m1 before and after, as well as `p3.M  = 0`.
What do you mean by `p3.M / p3.E = 0`.

blinkseb:
Awesome :+1: ! Once we are all in favor for merging, I'll run `clang-format` once to reformat the code (there's a lot of spaces missing around) and then push the green button.

Thanks again!

swertz:
Great thanks, the tests are now passing :+1:

I meant doing like that: https://github.com/MoMEMta/MoMEMta/blob/master/tests/unit_tests/modules.cc#L246
I think it makes more sense to check that the mass is zero with a precision relative to the particle's energy (when `E` and `p` are large, computing `E^2 - p^2` will never be precisely zero)...?

swertz:
:+1: for me, @blinkseb it's all yours ;)

blinkseb:
Thanks a lot! I merged this PR manually, see f44d80242e9cb6f62fd5da0369f445517c063842.

PR #138, opened 2017-04-19
Refactor phase space generation tests

swertz:
As already discussed, this modifies the phase space volume/cross section tests to have fewer executables (only two: `phaseSpaceVolume.exe` and `crossSections.exe`).

The tests are chosen using arguments, e.g.:
`tests/phaseSpaceGeneration/phaseSpaceVolume.exe A BE G` will compute the PS volume in turn for Main Block A, Main Block B + secondary E, and Main block G.

I've also added a small function to generate random 4-vectors with a given mass, since these tests require "dummy" input vectors (but with a definite mass). The loop in the function is there to avoid numerical issues where the mass gets slightly negative.

Also included is a PS volume test for Main Block E. There are still some tests missing (cross sections); I'll add them later. I'll also add more unit tests for blocks in a separate PR.


tests/phaseSpaceGeneration/phaseSpaceGenerationTests.cc
```diff
@@ -0,0 +1,71 @@
...
+        secondaryBlocks.emplace(std::make_pair(mainBlock, secondaryBlock), test);
+}
+
+void PSGenerationTestManager::parseArgs(int argc, char** argv) {
+    if (argc <= 1)
+        throw std::invalid_argument("Expected at least one argument!");
+
+    for (int i = 1; i < argc; i++) {
+        std::string arg(argv[i]);
+        if (arg.size() == 1)
```
blinkseb: `length()` instead of `size()` is maybe more common (I didn't know `size()` existed for `std::string`!)

include/momemta/Utils.h
```diff
@@ -70,6 +74,16 @@ void apply_permutations(std::vector<T>& vec, std::vector<std::size_t> const& p)
...
 
+/*!
+ * Generate a random Lorentz vector
+ *
+ * @m The mass of the vector (default 0)
+ * @maxE The maximum energy of the vector
+ *
+ * @return The random LorentzVector
+ */
+LorentzVector getRandom4Vector(double maxE, double m=0);
```
blinkseb: If this function is also intended to be used by the end-user, I'd wrap it inside a `momemta` namespace to avoid potential clash. What do you think?

tests/phaseSpaceGeneration/phaseSpaceGenerationTests.cc
```diff
@@ -0,0 +1,71 @@
+#include <momemta/ConfigurationReader.h>
```
blinkseb: Please add licence info

tests/phaseSpaceGeneration/phaseSpaceGenerationTests.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
```
blinkseb: Please add licence info

tests/phaseSpaceGeneration/phaseSpaceGenerationTests.cc
```diff
@@ -0,0 +1,71 @@
...
+#include <momemta/Types.h>
+#include <momemta/Utils.h>
+
+#include <chrono>
+#include <exception>
+
+using namespace std::chrono;
+using namespace momemta;
+
+#include "phaseSpaceGenerationTests.h"
```
blinkseb: This one should be the first header to be included

include/momemta/Utils.h
swertz: Hm fine with me, but shouldn't we then put all the functions in that file into the namespace?

tests/phaseSpaceGeneration/crossSections.cc
```diff
@@ -0,0 +1,51 @@
...
+#define XS_TT_DILEP 6.73 // +- 0.02
+
+int main(int argc, char** argv) {
+
+    logging::set_level(logging::level::debug);
+
+    PSGenerationTestManager manager;
+
+    // W(lnu)W(lnu)
+    manager.registerTest({ "../tests/phaseSpaceGeneration/crossSections/blockA_WW_dilep.lua", { "p1", "p2", "p3", "p4" }, { 0, 0, 0, 0 }, CONVERSION * XS_WW_DILEP }, 'A');
```
swertz: I'm not clear how to make that work... Naively doing what you're suggesting (with `target_compile_definitions(test_XS BASE_PATH=${PROJECT_BINARY_DIR})` doesn't work.
Would you suggest making the .cc file a template and use the `configure_file` command?

blinkseb: Why does it not work? That was my idea, something like
```
target_compile_definitions(test_XS PRIVATE "BASE_PATH=${PROJECT_SOURCE_DIR}")
```

swertz: Well I get errors like
```
<command-line>:0:12: erreur: expected primary-expression before ‘/’ token
/home/fynu/swertz/tests_MEM/MoMEMta/tests/phaseSpaceGeneration/crossSections.cc:36:28: note: in expansion of macro ‘BUILD_PATH’
     manager.registerTest({ BUILD_PATH "tests/phaseSpaceGeneration/crossSections/blockA_WW_dilep.lua", { "p1", "p2", "p3", "p4" }, { 0, 0, 0, 0 }, CONVERSION * XS_WW_DILEP }, 'A');
                            ^
<command-line>:0:13: erreur: ‘home’ was not declared in this scope
/home/fynu/swertz/tests_MEM/MoMEMta/tests/phaseSpaceGeneration/crossSections.cc:36:28: note: in expansion of macro ‘BUILD_PATH’
     manager.registerTest({ BUILD_PATH "tests/phaseSpaceGeneration/crossSections/blockA_WW_dilep.lua", { "p1", "p2", "p3", "p4" }, { 0, 0, 0, 0 }, CONVERSION * XS_WW_DILEP }, 'A');
```
It seems the variables get substituted by the path but it doesn't merge the path with the string in the .cc file (and I don't see why it should?)...

blinkseb: Ah it's because you need to add quotes to the definitions, like `BASE_PATH="${PROJECT_SOURCE_DIR}"`.

String merging is a feature of the language itself, doing
```
printf("Hello"
" world");
```

would print `Hello world`

swertz: Ah well that was rather obvious...

swertz:
Thanks for the comments! I think I have addressed them all.

blinkseb:
Indeed it looks perfect! Travis will fail unfortunately, I'll play around with this tonight and merge afterwards! Thanks a bunch!!

swertz:
Test is now passing, have you fixed Travis? :)

blinkseb:
I just ping hepforge yesterday to fix their SSL certificate ;) I checkout 1h ago and it was fixed so I retriggered Travis manually :)

blinkseb:
Thanks a lot again!

swertz:
Ok, thanks :)

PR #139, opened 2017-04-19
Add version check to FindLHAPDF.cmake

fnand:
I don't know if anyone else has this problem, but I cannot compile unless the LHAPDF version check is called. 

The fix is as simple as adding: 

`exec_program(${LHAPDF_CONFIG_PROGRAM} ARGS "--version" OUTPUT_VARIABLE LHAPDF_HAVE_VERSION)`

to cmake/modules/FindLHAPDF.cmake  

-Ferdi


swertz:
I'm sorry, none of us has ever had problems with LHAPDF... Maybe there's an issue with your installation?

Travis failed but I'm kind of puzzled as to why this innocent looking addition broke it... @blinkseb do you have an idea?

blinkseb:
Travis failed because hepforge.com SSL certificat is no longer valid (`ERROR: cannot verify www.hepforge.org's certificate, issued by ‘/C=US/O=Let's Encrypt/CN=Let's Encrypt Authority X3’:
  Issued certificate has expired.`), so it's not directly related to this PR.

Concerning the issue here, I'm not sure to understand why it's needed. Can you provide more details on the error you see without this change? The version string should be extracted automatically from the `Version.h` header.

fnand:
Hi, well, I am sure that it is not this update that causes issues as it doesn't do anything funny. 

But after some further investigation I found this might just have been a workaround and some of my env variables might have been screwy. 

So ultimately this isn't really neccesary... Sorry for that. Closing the PR

Issue #140, opened 2017-04-19
Unit tests `Approx` approximate a bit too much


blinkseb:
Using `Approx` in unit tests seems to approximate a bit too much. We should explicitly use the `epsilon` function to state how many significant figure we want to consider in the comparison (`Approx(x).epsilon(y)`)

PR #141, opened 2017-04-23
Add unit tests for all blocks (except main block C)

swertz:
Title says all!

It's been quite a pain to find dummy values for which the blocks actually returned solutions that could be tested... There might be a better way of doing that, to also cover the cases where the blocks are supposed not to return any solutions.

Regarding #140: I've left the plain `Approx` for now... Problem is, depending on the block, on the block inputs, on the value that is tested (`sX`, `M()`, `Pt()`), the precision varies quite a lot. 
We could try to find a common minimum value that works everywhere and would be sufficiently small to spot any problems (which is the point of those tests), but isn't the defaut value good enough for that purpose?

### DIFF
--- a/include/momemta/ParameterSet.h
+++ b/include/momemta/ParameterSet.h
@@ -135,9 +135,25 @@ class ParameterSet {
                                 std::is_same<T, InputTag>::value>::type set(const std::string& name, const T& value) {
             set_helper(name, value);
         }
+        
+        /**
+         * \brief Change the value of a given parameter. If the parameter does not exist, it's first created.
+         *
+         * Partial specialization for std::vector.
+         *
+         * \param name The name of the parameter to change
+         * \param value The new value of the parameter
+         */
         template<typename T>
-        typename std::enable_if<std::is_same<T, bool>::value ||
-                                std::is_same<T, InputTag>::value>::type set(const std::string& name, const std::vector<T>& value) {
+        void set(const std::string& name, const std::vector<T>& value) {
+            static_assert(
+                    std::is_same<T, int64_t>::value ||
+                    std::is_same<T, double>::value ||
+                    std::is_same<T, bool>::value ||
+                    std::is_same<T, std::string>::value ||
+                    std::is_same<T, InputTag>::value,
+                    "Type not supported"
+            );
             set_helper(name, value);
         }
 
@@ -148,8 +164,6 @@ class ParameterSet {
          *
          * \param name The name of the parameter to change
          * \param value The new value of the parameter
-         *
-         * \warning Vectors are currently not supported
          */
         template<typename T>
         typename std::enable_if<is_string<T>::value>::type set(const std::string& name, const T& value) {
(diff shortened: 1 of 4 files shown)


swertz:
Might be easier to merge this after #137 is over. I'll rebase on top. @fnand if you want you case re-use the massless 4-vectors as defined here https://github.com/MoMEMta/MoMEMta/pull/141/files#diff-5bd08e012f7e229f727cb63117110d0bR81 for your test!

include/momemta/ParameterSet.h
```diff
@@ -129,14 +129,17 @@ class ParameterSet {
...
         template<typename T>
         typename std::enable_if<std::is_same<T, bool>::value ||
                                 std::is_same<T, InputTag>::value>::type set(const std::string& name, const T& value) {
             set_helper(name, value);
         }
+        template<typename T>
+        typename std::enable_if<std::is_same<T, bool>::value ||
+                                std::is_same<T, InputTag>::value>::type set(const std::string& name, const std::vector<T>& value) {
+            set_helper(name, value);
+        }
```
blinkseb: this only allows `std::vector<bool>` and `std::vector<InputTag>`, is this intended? IMO maybe it's easier here to remove the `enable_if` and add a `static_assert` checking that `T` is a supported type, exactly like `set_helper` is doing.

swertz: I only did that one because I only needed it for InputTag vectors...

About the others: we can do it like that, but then what about the casts to the supported types, as in https://github.com/MoMEMta/MoMEMta/pull/141/files/bba7b7b28d9b92b5829213ec19d8c9207cdbaa47#diff-e170c8796893b6deb367b072aaa6aaa0R171 ?
I can keep those and just modify this one, which means the user has to explicitly use `std::vector<int64_t>` and `std::vector<double>`?

blinkseb: I mean doing that only for the `std::vector` overload, but keeping the others for basic types. There's no real issue here as you don't have implicit conversion between vectors: a `vector<double>` can't be implicitly converted to any other thing, and yes, the user have to correctly specify the type of the vector, there's no other way around (it's the same thing for the basic type, we just support implicit conversion from float to double and int to int64 basically)

swertz: That doesn't seem to work: if I remove the `enable_if`s, I cannot overload `set` with the different specialisations...

swertz:
Thanks for the advice, like that seems to work!

blinkseb:
Perfect!! Thanks a lot! One last thing now is to remove the comments saying that `vectors` are not supported :stuck_out_tongue:

swertz:
Not sure how to handle the different overloads/specializations for the documentation...? The specializations with cast to `double` and such still do not support vectors. Since you seemed to prefer having one doc entry per specialization, I thought keeping the two versions (vector allowed or not) separate would be better?

blinkseb:
That's the all points of specialization, you'll always call the `vector` overload if you use a vector, so with your changes, vectors *are* supported. I added the warning back when this was not supported, so it's safe to remove it. You can maybe add a line in the documentation of the vector overload stating something like `Partial specialization for std::vector`.

swertz:
Happy with that? :p

blinkseb:
Yes :p thanks a lot!

>  Coverage increased (+16.5%) to 72.444%

Nice :tada: !!!

PR #142, opened 2017-04-24
Refactor Travis configuration and add build with clang

blinkseb:
I started this a while ago and somehow forgot to PR it. This simply refactor a bit the travis configuration and adds a build using clang and more recent gcc version (5 and 6). I also took the opportunity to change two things:
  - In light of the hepforge breakage last week, I disable the SSL certificate check when downloading the PDF. This should prevent the same thing to happen in 3 months when they'll forgot again to renew the certificate :)
  - I also fixed a warning issued when compiling with clang

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,47 +1,161 @@
-os: [linux]
-sudo: required
+os: linux
 dist: trusty
-
 language: cpp
-compiler:
-  - gcc
-env:
-  global:
-    - PYTHON=true
-  matrix:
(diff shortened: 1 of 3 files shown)


swertz:
Thanks! Do we actually need the coverage on all the different builds?

blinkseb:
Probably not indeed, it was the previous behaviour. I'll enable it only on the main build.

swertz:
Ok, thanks again!

PR #143, opened 2017-04-28
(Partially) validate main block C

swertz:
Block C is now validated in what regards the phase-space volume check. There was an issue with the jacobian; now fixed.

I've also fixed a few typos in the documentation for block C.


blinkseb:
Perfect thanks a lot! Do you have xsec running by chance?

swertz:
No I'll do the xsec later, along with the other blocks! In principle the phase space volume should really be enough, I don't see how that test could succeed and the xsec fail, but if we want to double-check we'll want to do both...

PR #145, opened 2017-05-04
Cuba: always free shared memory

### LINKED ISSUE #144: Severe memory leak when killing or aborting the integration (cuba)
When the call to the Cuba integration routine does not end properly, the memory allocated by Cuba for the integration grid is not freed by the OS. I have not checked yet if it was the case when raising an exception during the integration.

The problem most probably comes from Cuba, but this might make us consider switching to another integration library?

swertz:
Fixes #144 

Cuba uses shared system memory (`shmget()`) to communicate with workers when using parallelisation. We were correctly asking that Cuba manages that memory itself and frees it when the integration is over. Problem is, when the program crashes, interrupts or is killed, that shared memory is not freed!

One solution is to manually check the shared memory entries using `ipcs` and remove them using `ipcrm`, but that's not quite user-friendly. This PR modifies Cuba to mark the shared memory page to be freed by the system as soon as no process is attached to it (which is the case even if the program is killed). This has to be done right after the call to `shmat()`, since that would fail if it is called on a page marked for removal.

I've checked that nothing bad happens when Cuba calls `shmctl()` again to free the memory when the call finishes properly.

### DIFF
--- a/core/src/MoMEMta.cc
+++ b/core/src/MoMEMta.cc
@@ -244,7 +244,7 @@ std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector
                     batch_size,             // (int) batch size for sampling
                     grid_number,            // (int) grid number, 1-10 => up to 10 grids can be stored, and re-used for other integrands (provided they are not too different)
                     grid_file.c_str(),      // (char*) name of state file => state can be stored and retrieved for further refinement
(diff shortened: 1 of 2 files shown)


vidalm:
Thanks!!

blinkseb:
Thanks a lot, nice catch! From the documentation, I read:

> Linux permits a process to attach (shmat(2)) a shared memory segment that has already been marked for deletion using shmctl(IPC_RMID). This feature is not available on other UNIX implementations; portable applications should avoid relying on it.

Something to keep in mind if we want in the future to support macOS or other Unix around

PR #147, opened 2017-05-09
Keep track of changes done to cuba

blinkseb:
This PR adds a set of patches of all the changes we've done to cuba, in order to easily reapply everything if ever a new version of cuba is released.


swertz:
Thanks! I was wondering if it'd be useful to actually fork Cuba into a separate directory, to apply the changes there... Probably not needed now but to keep in mind if there are other things that we want to change (simplify the interface).

blinkseb:
Since cuba itself is not versioned, I'm not sure how it'll simplify the things. You'd still have to overwrite all the files and then reapply patches on top.

PR #148, opened 2017-05-10
[WIP] Several minor fixes

swertz:
Several small changes, feel free to add stuff before merging this (I'll be offline for 10 days):

- Fixed typos in the doc
- Modified the conditions checked at the beginning of the blocks: some were redundant, some could be strenghtened (for instance, for s12->p1+p2, the masses of p1 and p2 put a lower limit on s12). That's not supposed to change the results but can make the blocks slightly faster (no need to do all the computations to discover at the end that there is no solution to the equations)
- Some modules were not in the docs' modules group. But I'm not sure what's the best way to handle the templated modules whose specializations have different names...
- I've attempted to validate SB A, but it's quite moody... I've only got ~98% of the correct value, and with a large number of points! More worryingly, repeating the computation many times with different seeds (and with a lower number of points), the integration always converges to about only ~90% of the correct value. Feel free to play around more!

### DIFF
--- a/modules/SecondaryBlockE.cc
+++ b/modules/SecondaryBlockE.cc
@@ -90,15 +90,16 @@ class SecondaryBlockE: public Module {
         virtual Status work() override {
 
             solutions->clear();
-
-            // Don't spend time on unphysical part of phase-space
-            if (*s12 > SQ(sqrt_s) || *s123 > SQ(sqrt_s) || *s12 > *s123)
-               return Status::NEXT;
-
+            
             const double m1 = m_p1->M();
             const double sq_m1 = SQ(m1);
-            const double m2 = m_p2->M();
             const double m3 = m_p3->M();
+            const double sq_m3 = SQ(m3);
+
+            // Don't spend time on unphysical part of phase-space
+            if (*s123 >= SQ(sqrt_s) || *s12 + sq_m3 >= *s123 || sq_m1 >= *s12)
+               return Status::NEXT;
+
             const double p3 = m_p3->P();
             const double E3 = m_p3->E();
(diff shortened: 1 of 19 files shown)


modules/SecondaryBlockE.cc
```diff
@@ -108,11 +109,11 @@ class SecondaryBlockE: public Module {
             const double c23 = ROOT::Math::VectorUtil::CosTheta(*m_p2, *m_p3);
 
             double X = p3 * c23 - E3;
-            double Y = *s123 - *s12 - SQ(m3);
+            double Y = *s123 - *s12 - sq_m3;
 
             std::vector<double> abs_p1, abs_p2;
             solve2Quads(SQ(X), SQ(p3 * c13) - sq_E3, 2 * p3 * c13 * X,  X * Y, p3 * c13 * Y, 0.25 * SQ(Y) - sq_E3 * sq_m1,
-                        2 * X / E3, 0, 2 * (p3 * c13 / E3 - c12), Y / E3, 0, sq_m1 + SQ(m2) - *s12,
+                        2 * X / E3, 0, 2 * (p3 * c13 / E3 - c12), Y / E3, 0, sq_m1 - *s12,
```
blinkseb: Is the removal of `m2` expected?

swertz: The block supposes `m2=0`. `p2` is supplied by the user so if that's not the case, the result will be rubbish, but that was already the case before!

blinkseb: Ok makes sense 👍

swertz:
Closing this since included in #149

PR #149, opened 2017-05-12
Several minor fixes

blinkseb:
This is #148 with an additional patch for cuba. For some reason I'm not able to edit the branch from Seb directly, so I open a new PR.


swertz:
Thanks for the addition!

PR #150, opened 2017-05-19
Fix compatibility with recent Doxygen versions

blinkseb:
Recenr Doxygen version changed the way the top menu is handled, breaking the custom theme we are using. This PR adapts the code to be compatible with these new releases. As a consequence, we now require at least Doxygen 1.8.12. Travis doxygen version has been updated to the latest one (1.8.13).


swertz:
Thanks!

PR #152, opened 2017-06-22
Fix spurious h2 in doxygen doc

blinkseb:
See https://momemta.github.io/MoMEMta/dev/classModule.html#a43c0b7c70e9dd8b0cf6c78aa5052eeee for an example of what this fixes (the big work with the small diamond on its left).


swertz:
Ah good forgot about that :+1:

Issue #155, opened 2017-07-25
Matrix element returning NaN will fail all following events


swertz:
Note: related to https://github.com/MoMEMta/MoMEMta-MaGMEE

Once the matrix element returns a NaN (e.g. evaluated on a particle of zero energy when soft divergences are present), it will always return zero in following calls, due to the selection of non-zero helicities.

While this is not an issue when computing a weight on a single event, this means that all following computations on other events will always return a zero integrand, and hence zero weights.

Two things are to be done:
- Protect the integrand returned by MoMEMta to Cuba to check for non-finite values
- Re-initialise the ME when starting a new event. This requires updating the exporter.

swertz:
Closed by #156

PR #156, opened 2017-07-26
Protect against NaN or inf integrands; reset ME at start of each integration

swertz:
- Protect against NaN of inf integrands: throw exception when integrand is non-finite. I'm actually not sure about this: maybe want not to crash the program but return a specific failure code for the integration? I think it'd be more easily missed...
- Reset ME at start of each integration: goes with update in the exporter


swertz:
Fixes #155

blinkseb:
Thanks a bunch! I think I prefer to throw an exception rather than returning a special code: a NaN is very bad, as it implies a bug somewhere, so this should not be un-noticed!

Let's see if travis is happy!

swertz:
It should work now, I had forgotten to update the particle PIDs in all the config files using the ttbar ME (I inadvertedly changed from mu+/e- to mu+/mu- final state, and was too lazy to re-generate the ME).

blinkseb:
Good to see our tests seem at least to be useful 👍

swertz:
I always forget to run the tests without pushing :D It seems fine now!

blinkseb:
Thanks a lot!

PR #157, opened 2017-08-30
Ensure block solutions are valid

blinkseb:
Algorithmes we use to solve the blocks' systems are not always numerically stable. To avoid potential issues resulting from invalid solutions (particles with negative mass, this kind of things), we know check each solution to ensure it's valid. If not, we simply throw it: this is much easier to do than trying to ensure all the solvers are numerically stables for every possible cases (I'm not sure it's even possible...).


modules/BlockB.cc
```diff
@@ -174,12 +174,35 @@ class BlockB: public Module {
...
+                if (!ApproxComparison(p1.M() / p1.E(), m1 / p1.E())) {
+#ifndef NDEBUG
+                    LOG(trace) << "[BlockB] Throwing solution because of invalid mass. " <<
+                               "Expected " << m1 << ", got " << p1.M();
+#endif
                     continue;
- 
+                }
+
+                if (!ApproxComparison((p1 + *p2).M2(), *s12)) {
```
swertz: I think it should be `p1 + pT`

modules/BlockC.cc
```diff
@@ -239,6 +239,38 @@ class BlockC : public Module {
...
 
+            if (!ApproxComparison((p1 + p3_sol + pT).Pt(), 0.)) {
+#ifndef NDEBUG
+                LOG(trace) << "[BlockC] Throwing solution because total Pt is incorrect. "
+                           << "Expected " << 0. << ", got " << (p1 + p3_sol + pT).Pt();
+#endif
+                continue;
+            }
+
+            if (!ApproxComparison(p1.M(), m1)) {
```
swertz: Above you did `p1.M() / p1.E()`. I'm sure it'll change anything, but maybe do it here also for consistency?

swertz: Below you also only didn't divide by the energy, I'll also not do it for the secondaries...

blinkseb:
PR updated, thanks for the comments!

PR #158, opened 2017-08-30
Check secondary block solutions are valid

swertz:
Same as #157 but for secondary blocks.

Note: it won't compile since I first have to rebase on #157, but at least the code can be reviewed.


swertz:
Also: some blocks reconstruct 4-vectors using input directions (theta and/or phi). I don't check the reconstructed output's angles matches those of the input, since that is trivially true when looking at how the outputs are defined.

blinkseb:
Thanks!

PR #160, opened 2017-10-20
Fix Block B

swertz:
The checks introduced in #157 for blockB were buggy (my bad, I was the one who requested that particular change).

Also for BlockB, the check for zero total transverse momentum is removed since it's trivially verified from the way the missing particle is reconstructed.

I've also added the PS volume validation for BlockE (no idea why it wasn't there, the value is in agreement with the analytical one) and fixed the reference value for BlockG.


swertz:
Once it's merged in master I'll push manually to the v1.0 branch...

blinkseb:
Let's go, thanks! 👍

Issue #161, opened 2017-11-03
Make the 'MoMEMta::integrand' function public


swertz:
In principle nothing prevents from making https://github.com/MoMEMta/MoMEMta/blob/master/core/src/MoMEMta.cc#L409 public. 

This would allows users to call the integrand configured using the whole machinery and the available modules, but do custom things such as using alternate integration algorithms, or something else entirely.

I don't see why it shouldn't work -> to be quickly tested some time.

delaere:
We can close this, right?

swertz:
Right, I thought it would be automatic through the reference from the PR...

PR #162, opened 2017-11-07
Evaluate integrand on single point

swertz:
As mentioned in #161, this adds the possiblity to call the integrand on a single phase-space point.

To care for the possibility where the user would like to use a custom integrator, I've added a function to set the event 4-momenta beforehand and only pass the phase-space point when computing the integrand.

Note: I'm having some issue with the python bindings on the UCL machines, let's see what Travis thinks of it here.

### DIFF
--- a/tests/integration_tests/CMakeLists.txt
+++ b/tests/integration_tests/CMakeLists.txt
@@ -1,4 +1,5 @@
 set(SOURCES
+    "integrand.cc"
     "no_integration.cc"
     "integration_tests.cc"
     )
--- a/core/src/MoMEMta.cc
+++ b/core/src/MoMEMta.cc
@@ -153,9 +153,9 @@ MoMEMta::~MoMEMta() {
 const Pool& MoMEMta::getPool() const {
     return *m_pool;
 }
+        
 
-std::vector<std::pair<double, double>> MoMEMta::computeWeights(const std::vector<momemta::Particle>& particles, const LorentzVector& met) {
-
+void MoMEMta::setEvent(const std::vector<momemta::Particle>& particles, const LorentzVector& met) {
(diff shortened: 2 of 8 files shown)


swertz:
FYI the tests pass, it's just test coverage that decreases. But OK I'll add a test for this new feature as well and update the PR.

swertz:
I've now added an "integration" test.

Fixes #161

blinkseb:
Looks like your forgot to add one file :smile:

swertz:
Not one, two ^^

swertz:
Wow, I have no idea what fails the tests... It runs absolutely fine on ingrid, but I get a segfault from the graph on my laptop... Investigating.

tests/integration_tests/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 set(SOURCES
+    "integrand.cc"
```
swertz: Ok, this makes no sense to me. If I put `integrand.cc` *after* `no_integration.cc`, I get weird things like segfaults and exceptions from the core momemta code (graph, ...). However in this order it works fine on my laptop. @blinkseb does it ring any bell?

swertz: How about it, the compiler version affects the order in which the files after put: in the present order it works with GCC 7.3 but not 4.9 (and vice-versa).

swertz: ... and of course, the tests use GCC 4.9 but seem to work.

swertz:
Rebased on top of #169 since it fixes the bug that appeared sporadically here, please merge that one first.

vidalm:
Good to me, thanks!

PR #163, opened 2018-03-06
Fix build with GCC7

swertz:
I'm not quite clear why it worked before, since `<functional>` is c++11 and we've always required the standard...


swertz:
I've updated this taking into account the issue raised in #165. It's only a workaround but it seems to solve the problem.

Let's see how the tests run...

include/momemta/ModuleRegistry.h
```diff
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <mutex>
+#include <functional>
```
delaere: don't forget the  `#include <numeric>` in modules/MatrixElement.cc

CMakeLists.txt
```diff
@@ -38,13 +38,17 @@ if (NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE RelWithDebInfo)
 endif(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
 
-# Compiler requirement
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF) # Stick to the standard
+# Require c++11 *at least*, use default compiler standard if possible
+if (CMAKE_CXX_STANDARD_COMPUTED_DEFAULT LESS 11)
```
delaere: I think that `CMAKE_CXX_STANDARD_COMPUTED_DEFAULT` is a string. So you should use `STRLESS "11"` or 
`VERSION_LESS "11"`

swertz: Good point! Sadly, none of these work. The issue is that C++98 is not recognised as being older than C++11 due to 98 being larger than 11 :(
There's another way, let me try that...

include/momemta/ModuleRegistry.h
swertz: that had been done...?

swertz:
Damn it, it still doesn't work with ROOT built with c++14... CMake still adds the `-std=c++11` flag.

swertz:
By the way, we probably need a new test case for GCC7 and with ROOT built with c++14...

blinkseb:
> By the way, we probably need a new test case for GCC7 and with ROOT built with c++14.

That's a bit of work. I used a vagrant image of Ubuntu 14.04 to build ROOT with the proper libraries. You'll need to to the same with GCC 7 (if it's available easily in Travis, but it seems to be the case: ubuntu-toolchain-r-test), enabling C++14 and then add a new matrix configuration for Travis.

For the current issue, I see two solutions:

  1. Do something like
```


include(CheckCXXCompilerFlag)

# Check for standard to use
check_cxx_compiler_flag(-std=c++17 HAVE_FLAG_STD_CXX17)
if(HAVE_FLAG_STD_CXX17)
    # Have -std=c++17, use it
else()
    check_cxx_compiler_flag(-std=c++1z HAVE_FLAG_STD_CXX1Z)
    if(HAVE_FLAG_STD_CXX1Z)
        # Have -std=c++1z, use it
    else()
        # And so on and on...
    endif()
endif()
```
ugly but working

  2. Add a flag like ROOT is doing, `ENABLE_CXX14` to turn on C++ 14 support instead of C++ 11.

swertz:
> That's a bit of work

Yup, that's why it's on the todo list :D

> 1. Do something like (...) ugly but working

Right, that's what I was hoping to avoid, especially since all the `c++1x/y/z` flags are not really standard.

> 2. Add a flag like ROOT is doing

Thanks, that'll do it indeed... I'll try one more thing and revert to doing that if it doesn't work.

swertz:
The current thing works but not with CMake 3.2 (that's the test case failing travis)...

So, question: do we bump the cmake requirements or use @blinkseb's second proposal?

blinkseb:
I'd vote to bump cmake version personnaly.

(I was googling for `CMAKE_CXX_STANDARD_COMPUTED_DEFAULT` and found this thread https://stackoverflow.com/questions/49231035/conflict-on-c-standard-with-cmake :joy: Google is so fast at indexing things, it's crazy)

swertz:
I assume stackoverlow is prioritized :D

One of the proposed solutions might work: add a few of the c++11 features that we do use, such as `cxx_variadic_templates`, `cxx_generalized_initializers`, `cxx_lambdas`, ... That way cmake should stick to the default compiler option even if it is more recent than 11.

Note to myself: the exporter also needs updating https://github.com/MoMEMta/MoMEMta-MaGMEE/blob/master/Template/CMakeLists.txt#L30
And the tutorials: https://github.com/MoMEMta/Tutorials/blob/master/CMakeLists.txt#L21

include/momemta/ModuleRegistry.h
delaere: indeed, sorry. I missed it.

swertz:
This works here but doesn't compile with my c++14 ROOT; cmake still imposes the c++11 flag. I'll wait for a response to that on SO before reverting the latest commit and bump the cmake requirements.

swertz:
Updated and rebased. We'll now require cmake 3.4.0. It's not the perfect solution but it seems to work (huzzah).

vidalm:
So, it works... Blind merging from my side. Whatever is inside cmake is like Chinese for me.

PR #164, opened 2018-03-08
Licensing

delaere:
This PR adds additional information needed for the GPLv3 license. 
It also adds missing information about the copyright.

This has been prepared in collaboration with the UCL LTTO (Matthieu Palate).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,11 @@
+MoMEMta software has been developped by:
+=====================================
+
+Sébastien Brochet (Universite catholique de Louvain)
+Sébastien Wertz (Universite catholique de Louvain)
+Miguel Vidal (Universite catholique de Louvain)
+Brieuc François (Universite catholique de Louvain)
+Alessia Saggio (Universite catholique de Louvain)
+
+For their suggestions, the authors are indebted to:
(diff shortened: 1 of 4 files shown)


swertz:
Thanks, good to get that right! It'd be good to have that backported to the v0.1 and v1.0 branch, I suppose.

I am surprised we have to include so much about ROOT, Lua and LHAPDF, given that we don't redistribute them (contrary to cuba)? Because there are more dependencies than that (python and boost to begin with)...

delaere:
Well, for LUA we do.

That said, there is a bit of grey zone there. We have to include/follow the licensing terms for all the dependencies , but there is an exception for the build system + standard libraries. So you don't have to mention python (language/compiler/interpreter), or boost (considered as a standard library). For ROOT and LHAPDF, linking to them is enough.

swertz:
Right, forgot about lua... Ok, thanks again!

swertz:
Something else I forgot about: we also ship the catch header-only library:
https://github.com/MoMEMta/MoMEMta/blob/master/tests/include/catch.hpp
https://github.com/catchorg/Catch2/

It's the boost license so there's no real issue there...

PR #165, opened 2018-03-08
Compatibility with gcc version 7.3 and ROOT 6.12

delaere:
Changes needed to compile on Fedora 27, with gcc version 7.3 and ROOT 6.12.

ROOT 6.12 requires C++14, but with C++ a couple of explicit std includes are needed. Not sure that we want to enforce C++14 instead of C++11...


swertz:
That is weird: on `ui1` with GCC7.3 and ROOT 6.12 I only needed the change in #163 to get it to work...

Adding the explicit includes is not a problem, but I don't want to enforce C++14... It is probably because you built your own ROOT with C++14 enabled?

It doesn't work if you only add the includes but don't touch the cmake script?

delaere:
without change, it complains at the first include of TFile that it requires C++14:
```
In file included from /usr/include/c++/7/experimental/string_view:39:0,
                 from /usr/include/root/RStringView.h:24,
                 from /usr/include/root/TString.h:28,
                 from /usr/include/root/TNamed.h:26,
                 from /usr/include/root/TDirectory.h:25,
                 from /usr/include/root/TDirectoryFile.h:24,
                 from /usr/include/root/TFile.h:26,
                 from /home/delaer/Tools/MoMEMta/modules/BinnedTransferFunctionOnEnergy.cc:26:
/usr/include/c++/7/bits/c++14_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2014 standard. This support must be enabled with the -std=c++14 or -std=gnu++14 compiler options.
 #error This file requires compiler and library support \
  ^~~~~
```

I use the ROOT version packaged by Fedora.
```
   ------------------------------------------------------------
  | Welcome to ROOT 6.12/06                http://root.cern.ch |
  |                               (c) 1995-2017, The ROOT Team |
  | Built for linuxx8664gcc                                    |
  | From tag v6-12-06, 9 February 2018                         |
  | Try '.help', '.demo', '.license', '.credits', '.quit'/'.q' |
   ------------------------------------------------------------
```

swertz:
Ok, we're not the first ones having this issue... cf. https://bugzilla.redhat.com/show_bug.cgi?id=1351623

The problem is that our cmake config really enforces c++11 and that there is no simple way of specifying "c++11 OR default IF default >= 11". Apparently the correct way to do it would be using [target_compile_features](https://cmake.org/cmake/help/latest/command/target_compile_features.html); it seems a bit cumbersome to set up but I can try something.

swertz:
Closing this since it's addressed by #163

PR #168, opened 2018-03-14
Review module definitions

swertz:
We missed a few details when setting up the nice new way of defining modules (#151). I've gone through it again, let's hope everything is correct now! Thanks for @fbury for spotting the bug.


vidalm:
Thanks!

PR #169, opened 2018-03-14
Fix bug in graph creation

swertz:
This is a naaasty bug that was plaguing me (see #162)... 

Here: https://github.com/MoMEMta/MoMEMta/blob/master/core/src/Graph.cc#L529 the lambda is declared static, but captures `this`. So as soon as the class is instanciated more than once, the pointer is ill-defined, leading to a nice Schrödinbug. This fixes it.

Another thing: since `graph_created` is set to `true` only at the end, the graph can not be exported is there is an issue in the config file. I've put it as soon as possible so that the graph can be inspected in case of crash.

Note: based on #163, please merge that one first.


vidalm:
Thanks again.

blinkseb:
Very nice catch !!

PR #170, opened 2018-03-15
add CATCH2 copyright and licence

delaere:


swertz:
Thanks! Actually, we're not using Catch v2 yet, there's been many updates since. While you're at it could you please update the catch file? It's just a matter of downloading https://raw.githubusercontent.com/CatchOrg/Catch2/master/single_include/catch.hpp

delaere:
Ok, I will do.

tests/unit_tests/modules.cc
```diff
@@ -130,7 +130,7 @@ TEST_CASE("Modules", "[modules]") {
 
         REQUIRE(module->work() == Module::Status::OK);
 
-        REQUIRE(*s == Approx(0));
+        REQUIRE(*s == Approx(0).margin(std::numeric_limits<float>::epsilon()));
```
swertz: Was that needed? I think at some point I tried updating catch and the only change needed to make it work what the #include (otherwise I would have made a separate PR and not asked you to add it here)...

delaere: yes, basically all the Approx(0) failed. Somebody raised the issue and the reason is that now the default is relative difference rather than absolute difference. Note that there are even places where epsilon is not enough. I put an absolute value by hand... maybe we can do better.

swertz: Ok, fair enough! These mass conditions are really difficult to get right, so it's ok...

swertz:
Thanks for taking care of it!

Issue #167, opened 2018-03-13
CMake doesn't find python 3.X


swertz:
On systems with both python2 and python3 installed (which is most of them), the current cmake config is not able to find python 3. Even forcing the minimum version to 3.0 doesn't work. 

Besides, because of [PEP394](https://www.python.org/dev/peps/pep-0394/), direct calls to `python`such as https://github.com/MoMEMta/MoMEMta/blob/master/tests/bindings/python/integration_tests.py#L1 will default to python2, which is not what we want.

Specifying the paths to python libs/interpreter/includes explicitly works, but the correct version of boost-python is not found, for the same reason (libboost_python.so links to the python2 version, not to (libboost_python3.so - at least on my 'buntu).

delaere:
googling around and experimenting, it really looks like a bug in cmake find_package(PythonLibs). Looks like they are preparing a separate find_package(Python3Libs) : https://gitlab.kitware.com/cmake/cmake/merge_requests/1819
we could maybe include these developments in our tree

swertz:
Thanks for the link! I've tried to naively download the updated modules from the PR, but there are too many changes for that to work. Even when that is merged, there's no chance we can support only the most recent cmake version...

Actually the problem is not so much python - setting the minimum version to 3 does work, contrary to what I write above - but finding the corresponding boost python.

I've managed to get something working using the changes proposed here (but not merged): https://gitlab.kitware.com/cmake/cmake/merge_requests/35/
Putting [the modified version of FindBoost.cmake](https://gitlab.kitware.com/cmake/cmake/raw/85a0a93ce18014a0f9f7314c97d7c888a717eee7/Modules/FindBoost.cmake) into our `cmake/modules` directory and running cmake with `-DBoost_PYTHON_VERSION_TAG=-py36` worked like a charm!

So, we could ship this file until a better solution becomes mainstream, and require the user to:
- Force the min. version of python to 3 (that could easily be an additional build option)
- Specify what is the suffix for the boost::python library, since that varies from system to system. On ubuntu finding that is as easy as running `locate boost_python`, but that might be less true on other systems...

Then there's the issue of the `python` call in the test, but since the script is already being configured by cmake (see [here](https://github.com/MoMEMta/MoMEMta/blob/master/cmake/scripts/run_tests.sh.in)), that can easily be solved.

swertz:
For further reference, another relevant CMake open issue is https://gitlab.kitware.com/cmake/cmake/issues/16391

[...]

PR #171, opened 2018-03-15
Handle both versions of python

swertz:
Attempt at fixing #167 (see the issue for a more complete description)

I still need to update the README - I'll do this if you think this is an acceptable solution.

It also needs to be tested under different environments (it works on my laptop with 'buntu). And it'd be good to have travis test cases with python3, too.


delaere:
I have an issue here... I tried on my machine and the version of boost is more recent, so I had to patch the findBoost code. Still, there are many difference between the system version and my patched version. We should find a way to do the same without replacing the full findBoost.cmake.

Maybe related, when I finally compile with the python3 bindings, I end up with the following error when running the tests:
>Traceback (most recent call last):
>  File "/home/delaer/Tools/MoMEMta/tests/bindings/python/integration_tests.py", line 31, in test_pyroot
>    p3 = momemta.Particle("electron", LorentzVector(16.171895980835, -13.7919054031372, -3.42997527122497, 21.5293197631836))
>Boost.Python.ArgumentError: Python argument types in
>    Particle.__init__(Particle, str, LorentzVector<ROOT::Math::PxPyPzE4D<double> >)
did not match C++ signature:
>    __init__(_object*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, ROOT::Math::LorentzVector<ROOT::Math::PxPyPzE4D<double> >, long)
>    __init__(_object*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, ROOT::Math::LorentzVector<ROOT::Math::PxPyPzE4D<double> >)
>    __init__(_object*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)

swertz:
And your PyROOT is built with the same python version?

delaere:
I think so.
Initially I was missing the python3-root package and cmake failed. But strangely, /usr/lib64/root/libPyROOT.so is the same for python3-root and python2-root and refers to libpython2.7.so.1.0 ! So maybe a packaging issue in Fedora.

delaere:
ok, the packaging issue is confirmed. I unsinstalled python2-root and reinstalled python3-root and it works. So python2-root  and python3-root cannot coexist on Fedora!

swertz:
I think that's the case everywhere, it's not possible to have pyroot for both versions of python (unless you build it twice, but the libraries have the same name).

Great that it worked, but there's another issue on ingrid: cmake doesn't find python 3 at all...

swertz:
Or, for some reason, when the order of `find_package(PythonInterp)` and `find_package(PythonLibs)` is inverted, it works.

There is still an issue in that boost::python is only available for python2 on ingrid, but that's not a problem with MoMEMta...

delaere:
We should still find a better solution for findBoost... maintaining it could become a nightmare, especially for the _Boost_COMPONENT_DEPENDENCIES function.

swertz:
You're right... Here's another attempt that seems to work on my laptop. The `Boost_PYTHON_VERSION_TAG` variable still has to be passed by the user, but I was able to use that tag to modify our own cmake config to use the default boost cmake module.

delaere:
Works for me too. I have a warning about boost headers, but I suppose that we can ignore it.

swertz:
I do too, but the build is fine... There's still a confusing issue with boost_python on ingrid (see https://cp3-git.irmp.ucl.ac.be/cp3-support/helpdesk/issues/1572), that has to be solved before I'm sure we can merge this.

swertz:
Ok, it seems to work fine now! I've updated the README for the new build options.

I only have one peculiar issue: when building with `-DPROFILING=ON` (with `gperftools` installed!), I the python unit test fails:
```
Error in `/usr/bin/python3': free(): invalid pointer: 0x00000000029290e0
```
Googling a bit I couldn't find much, but it seems that when linking with `gperftools`, the standard `malloc()` calls get replaced by something else, and that could be the source of the issue. 

It doesn't look easy to fix, but it's perhaps not urgent given that users will rarely want to profile anyway?

swertz:
I've found the issue with the profiling: two additional libraries are linked, `tcmalloc` for heap profiling & checking, and `profiler` for CPU profiling. The problem is, while the python module library is linked again `tcmalloc`, python itself is not, hence the mismatch...

The solution is then to do e.g. `LD_PRELOAD=(path to tcmalloc) ./run_tests.sh` to make sure `tcmalloc` is available for everyone. I'll update the PR.

swertz:
Updated again, and I hope for the last time!

vidalm:
Thanks a lot btw!!

Issue #166, opened 2018-03-12
Add test case with GCC7 and ROOT built with explicit c++14


swertz:

swertz:
Note: related to #159

PR #173, opened 2018-04-26
new path to travis build bundles

delaere:
copied all files from sbrochet to my CP3 public_html


swertz:
Thanks! We'll now be able to add tests with GCC7 and/or python3... (see #166)

PR #174, opened 2018-05-16
Fix finding LHAPDF

swertz:
FindLHAPDF.cmake was modified in #107 to support CMSSW, but it does not seem to work (anymore?)...

The workaround is needed in case `lhapdf-config` is not available, which is the case inside a CMSSW environment.

Tested on CMSSW_9_4_5.


swertz:
The failed travis build appears because I pushed a branch by mistake to this repository, and then deleted it - you can disregard it.

PR #175, opened 2018-06-29
Set lhapdf path from environment variable

swertz:
For environments where `LHAPDF_ROOT` is set as environment variable, this makes CMake find it without user intervention.

Note: please don't merge until I've checked with the people who reported the issue if that fixes it completely.


swertz:
Closing this... No news from the people from reported it. In any case, specifying the variable as `-DLHAPDF_ROOT=(path)` when calling `cmake` will always work.

PR #179, opened 2021-06-14
docs: Show how root-config can get ROOT's CMAKE_CXX_STANDARD value

matthewfeickert:
* `root-config --cflags` will output a string that includes '-std=c++X' where 'X' is the C++ standard version, and then sed with regex can be used to extract that number

Example:
```console
$ root-config --cflags
-pthread -std=c++17 -m64 -I/home/feickert/bin/root-cern/include
$ echo $(root-config --cflags | sed --regexp-extended 's/.*-std=c\+\+([0-9]+).*/\1/g')
17
```


matthewfeickert:
Thoughts on the use of this @swertz?

swertz:
Thanks, that's very useful!

Perhaps this could then be added directly into the cmake config instead of exposing these details in the docs? Using something like
```
execute_process(
    COMMAND root-config --cflags
    COMMAND sed --regexp-extended 's/.*-std=c\+\+([0-9]+).*/\1/g'
    OUTPUT_VARIABLE ROOT_CXX_STANDARD
    OUTPUT_STRIP_TRAILING_WHITESPACE)
```
and then using `${ROOT_CXX_STANDARD}` as the C++ standard for the MoMEMta build? (command not tested)

PR #181, opened 2023-04-25
New root features

jtchk:
`getSolutions()` will be a new python function which will return a python list of the solutions generated by a Block. 
The inputs are the name of the block, as specified in the Lua file, in the form of a python string and the python list of `Particle`'s of final state particles. The evaluation of the weight is necessary s the MET will be computed beforehand. 

Several changes are incorporated in order to make MoMEMta consistent with the new versions of ROOT (>=6.20.00).


core/src/python/bindings.cc
```diff
@@ -100,6 +101,48 @@ bp::list MoMEMta_computeWeights(MoMEMta& m, bp::list particles) {
     return MoMEMta_computeWeights_MET(m, particles, bp::list());
 }
 
+bp::list MoMEMta_getSolutions(MoMEMta& m, const std::string& blockName, bp::list particles_) {
+
+    /* Extract the missing transverse Energy */
+    bp::list met_;
```
swertz, line 107: I don't understand how the MET is handled.

For the regular particles, you retrieve them from the `particles_` python list. But the `met_` list is just created here, while I think it should also be passed as an argument to the function?

core/src/python/bindings.cc
```diff
@@ -100,6 +101,48 @@ bp::list MoMEMta_computeWeights(MoMEMta& m, bp::list particles) {
...
+    auto weights = m.computeWeights(particles, met);
+
+    /* Define the address of the solutions of the required block */
+    InputTag solInputTag {blockName, "solutions"};
+
+    /* Retrieve solutions out of the memory pool */
+    auto sol = m.getPool().get<SolutionCollection>(solInputTag);
+
+    /* Create the converter to create a Boost Python 4-momentum out of a LorentzVector */
+    /*LorentzVector_to_python four_momentum_converter;*/
```
swertz, line 129: Can you remove this if it isn't used?

swertz (changes requested):
Thanks!

Could you just please add a meaningful description for your pull request?

Can you remove the empty `build.log` file from the PR?

Can you also bump the minimally supported ROOT version to 6.20.00 at these two places:
- https://github.com/MoMEMta/MoMEMta/blob/master/CMakeLists.txt#L86
- https://github.com/MoMEMta/MoMEMta/blob/master/README.md

core/src/python/bindings.cc
jtchk: Lines 128-129 have been deleted from the `bindings.cc`file.

swertz:
Thanks again!

swertz (approved):
LGTM

